PR #174, opened 2026-04-30
[LOOP-166] strip stale pipeline labels on terminal state

svv2014:
Closes #166

## Changes
- **`lib/labels.sh`** (new): single source of truth for `LOOP_PIPELINE_STAGE_LABELS` and a `loop_strip_pipeline_labels` helper. Stage set: `plan, po-review, dev, in-progress, in-rework, in-review, review-pending, needs-review, needs-rework, needs-qa, needs-clarification, ready-for-qa, qa-pass, qa-fail, changes-requested`. Orthogonal labels (priority, `semver:*`, `epic`, `tracker`, `release-pr`, `safe-to-test`, `bug`, `feature`, `blocked`) are never touched.
- **`scripts/merge-handler.sh`**: after merge, strips every stage label from both the merged PR and each linked issue before applying `done` to the issue. Replaces the old single-label `qa-pass` removal.
- **`scanner/reconciler.sh`**: new `reconcile_closed_issue_labels` check walks issues closed in the last `LOOP_CLOSED_LOOKBACK_DAYS` (default 7) and strips any leftover stage labels. Idempotent — second tick finds nothing.
- **`scripts/backfill-stale-labels.sh`** (new): one-shot historical cleanup over merged PRs + closed issues from the last 30 days (`--days N` to override). Defaults to `--dry-run`; `--apply` to mutate. Honours `--slug` for single-project runs. Idempotent: second `--apply` pass strips nothing.
- **`tests/stale-labels.bats`** (new, 5 cases): scoped strip behavior, merge-handler post-merge sequence, reconciler closed-issue path, backfill idempotency.

All three call-sites read the stage set from `lib/labels.sh`, so future drift only happens in one place.

## Test Plan
- [ ] `bats tests/stale-labels.bats` — 5/5 pass
- [ ] `bash -n` and `shellcheck -S warning` clean on all touched files
- [ ] Dry-run the backfill on this repo and confirm the planned diffs match the examples in #166 (PR#150, PR#146, issue #141)
- [ ] Apply the backfill on this repo; re-run and verify zero changes

Issue #168, opened 2026-04-30
[LOOP] label vocab: reconciler renames deprecated labels on sight


svv2014:
## Objective
Reconciler scans every open issue/PR each tick; if a deprecated alias is present, swap it for the canonical label (additive then subtractive, never lossy). Idempotent.

## Acceptance Criteria
- [ ] New reconciler check (sibling to existing checks in `scanner/reconciler.sh`) iterates open issues+PRs per project
- [ ] For each ticket carrying any `LOOP_DEPRECATED_ALIAS_MAP` key: add the canonical label first, then remove the deprecated one
- [ ] Idempotent: a second invocation makes zero label changes
- [ ] Counter surfaced in reconciler summary log as `alias_renamed=N`
- [ ] Bats coverage: ticket with deprecated label gets renamed; ticket already canonical untouched

## Files & Scope
- Touch: `scanner/reconciler.sh`, `tests/*.bats`
- Do NOT touch: handlers under `scripts/`, backend adapters

## Dependencies
First child (lib/labels.sh canonical mapping) must land first.

## Notes
Parent epic: #165. Use the backend abstraction (`backend_add_label` / `backend_remove_label`) — never call `gh` directly per CLAUDE.md.

## Out of scope
- Deleting deprecated labels from the repo (separate child)
- Workflow YAML edits (separate child)

svv2014:
PO: spec already in canonical form. Queuing for implementation. Blocked on #167 (lib/labels.sh) shipping first.

svv2014:
Blocked by #167 (sibling child of epic #165): `lib/labels.sh` and `LOOP_DEPRECATED_ALIAS_MAP` do not yet exist in the repo. The acceptance criteria reference that map directly ("For each ticket carrying any LOOP_DEPRECATED_ALIAS_MAP key…") and the issue's own Dependencies section states: "First child (lib/labels.sh canonical mapping) must land first."

Searched the tree:
- `lib/labels.sh` — does not exist
- `LOOP_DEPRECATED_ALIAS_MAP` — zero matches

Holding implementation until #167 ships the canonical mapping; once that's merged, this reconciler check can source `lib/labels.sh` and iterate the map. Re-labeling `needs-clarification`.

Issue #167, opened 2026-04-30
[LOOP] label vocab: define canonical mapping in lib/labels.sh + migrate handlers


svv2014:
## Objective
Establish `lib/labels.sh` as the single source of truth for canonical loop label names + their deprecated aliases, and update every handler to read from it (no string literals).

## Acceptance Criteria
- [ ] `lib/labels.sh` exports `LOOP_CANONICAL_LABELS` (the ~11 canonical names) and `LOOP_DEPRECATED_ALIAS_MAP` (deprecated → canonical)
- [ ] Canonical set: `needs-po`, `in-po`, `needs-dev`, `in-dev`, `needs-review`, `in-review`, `needs-qa`, `qa-pass`, `qa-fail`, `blocked`, `done`
- [ ] Deprecated aliases mapped: `po-review`→`needs-po`, `dev`/`plan`/`in-progress`→`needs-dev`, `review-pending`→`needs-review`, `ready-for-qa`→`needs-qa`, `needs-rework`/`changes-requested`→`needs-dev`, `in-rework`→`in-dev`
- [ ] All handlers under `scripts/` and scanner under `scanner/` reference these constants — no remaining hardcoded loop-label strings (grep -rE 'po-review|review-pending|ready-for-qa|in-rework|changes-requested' scripts/ scanner/ returns nothing)
- [ ] Bats coverage that asserts every deprecated alias resolves to a canonical label

## Files & Scope
- Add: `lib/labels.sh`, `tests/labels.bats`
- Touch: `scripts/*.sh`, `scanner/scanner.sh` (only label-string replacements)
- Do NOT touch: `config/workflows/*.yaml` (separate child), `lib/backends/`, `config/projects.yaml`

## Dependencies
#166 (stale-label cleanup) should ship first so this migration runs over clean data.

## Notes
Parent epic: #165. Per CLAUDE.md, handlers source `lib/env.sh` / `lib/config.sh` / `lib/backends/backend.sh` — append `lib/labels.sh` to that pattern.

## Out of scope
- Renaming labels on the GitHub side (separate child)
- Workflow YAML edits (separate child)

svv2014:
PO: spec already in canonical form. Queuing for implementation.

svv2014:
🔁 Rework attempt 1/2 started (reviewer feedback) — PR #176 (2026-04-29 23:38)

svv2014:
[...]

Issue #165, opened 2026-04-30
[Epic] Consolidate label vocabulary — Symphony-inspired cleanup


svv2014:
## Context

Comparison with [OpenAI Symphony](https://github.com/openai/symphony) (released 2026-04-28) reveals loop's label vocabulary is overgrown. Symphony exposes only two operator-facing states (`Todo`, `In Progress`) and keeps its internal worker state machine private. Loop conflates the two and has accumulated synonyms for the same state.

Today's `gh label list` for svv2014/loop:

| Stage | Labels in use | Problem |
|---|---|---|
| Dev trigger | `dev`, `plan`, `in-progress` | 3 labels — unclear which the scanner triggers on |
| Review queue | `needs-review`, `review-pending` | Synonyms |
| Review in-flight | `in-review` | OK |
| Rework queue | `needs-rework`, `changes-requested` | Synonyms |
| Rework in-flight | `in-rework` | OK |
| QA queue | `needs-qa`, `ready-for-qa` | Synonyms |
| QA outcome | `qa-pass`, `qa-fail` | OK |
| Other | `po-review`, `blocked`, `done`, `needs-clarification`, `safe-to-test` | OK |

Two distinct problems:
1. **Operator-set vs agent-set labels are mixed** — handlers can race or disagree about which label "owns" a state.
2. **Multiple synonyms per state** — every duplicate is a place where `gh issue list --label X` misses tickets in state X-but-labelled-Y.

## Goal

Pick one canonical label per stage; demote duplicates to deprecated aliases that the reconciler renames on sight. Naming pattern (Symphony-inspired):

- `needs-{stage}` = queued for that stage (operator- or agent-set, idempotent)
- `in-{stage}` = worker has claimed the ticket (agent-only)
- terminal stages stay (`qa-pass`, `qa-fail`, `blocked`, `done`)

Proposed canonical set (~9 labels, down from 15):

```
needs-po, in-po, needs-dev, in-dev,
needs-review, in-review,
needs-qa, qa-pass, qa-fail,
blocked, done
```

Old labels (`po-review`, `dev`, `plan`, `review-pending`, `ready-for-qa`, `needs-rework`, `in-rework`, `changes-requested`, `in-progress`) become deprecated aliases. Reconciler maps them on sight; workflow YAMLs reference only canonical names.

## Deliverables (decompose into child tickets)

This epic should be split into sequential phases — **do not big-bang the rename**, since loop scans 8 projects with active tickets and labels are currently in flight.

[...]

PR #178, opened 2026-04-30
[LOOP] reconciler: rename synonym labels to live-workflow trigger names

svv2014:
## Summary

Adds `reconcile_synonym_labels` — a per-tick reconciler check that renames four deprecated synonyms agents accidentally apply to the live workflow trigger names.

## Why

The scanner only polls workflow-YAML trigger labels. When a review/qa/dev handler agent applies a synonym (`review-pending` instead of `needs-review`, `changes-requested` instead of `needs-rework`, etc.), the ticket falls out of the pipeline silently. **Today this bit PRs #148, #174 (and a series of others before that)** — operators had to hand-relabel.

Root cause: the agent prompts in handlers reference both names interchangeably, and `lib/labels.sh`'s canonical set is mid-migration (uses `needs-po`/`needs-dev`) while the live workflow YAML still triggers on `po-review`/`dev`/`needs-rework`. That broader unification belongs to #165 / #167-family. **This PR is the immediate stuck-ticket remedy**, narrow enough to ship while the vocab migration is in flight.

## Map

| from | to |
|---|---|
| `review-pending` | `needs-review` |
| `ready-for-qa` | `needs-qa` |
| `changes-requested` | `needs-rework` |
| `plan` | `dev` |

## Wire-up

`run_project()` calls `reconcile_synonym_labels "$REPO"` immediately after `reconcile_labels` (which ensures the canonical labels exist on the repo).

## Verification

- `bash -n scanner/reconciler.sh` — clean
- Dry-run against svv2014/loop produces no false positives (today's stuck items were hand-relabeled first).
- Idempotent: a second tick finds nothing.

## Follow-up

Bats coverage is deferred: `scanner/reconciler.sh` runs `acquire_lock` + main loop on source, which makes it un-sourceable from tests today. A small guard refactor (`[ "\${BASH_SOURCE[0]}" = "\$0" ] && main`) would make this and other reconciler functions unit-testable. Tracking via existing #168 (reconciler renames deprecated labels — child of #165 epic) which now has a concrete narrow scope.

### DIFF
--- a/scanner/reconciler.sh
+++ b/scanner/reconciler.sh
@@ -658,6 +658,81 @@ reconcile_labels() {
     log "[$repo] labels ok (created=$created)"
 }
 
+# --- Check 9b: rename synonym labels to live-workflow trigger names ----------
+# Some review/QA/handler agents apply synonym labels that are not the live
+# workflow trigger (e.g. `changes-requested` instead of `needs-rework`,
+# `review-pending` instead of `needs-review`). The scanner only polls workflow
+# triggers, so synonym-labelled tickets get stuck. Rename on sight.
+#
+# Map is intentionally narrow: only synonyms that can be safely promoted to
+# the live default-workflow trigger names. The broader vocab unification
+# (LOOP-167 / #165) is the canonical fix; this is the immediate stuck-ticket
+# remedy.
+LOOP_SYNONYM_MAP=(
+    "review-pending:needs-review"
+    "ready-for-qa:needs-qa"
+    "changes-requested:needs-rework"
+    "plan:dev"
+)
+
+_rename_label_on_target() {
+    # _rename_label_on_target <repo> <number> <kind> <from> <to>
+    # kind: issue | pr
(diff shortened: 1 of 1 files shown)


svv2014:
Reviewer (self-PR; cannot use formal approve): LGTM. Narrow stuck-ticket remedy that fits between live workflow triggers (`needs-review` / `needs-qa` / `needs-rework` / `dev`) and the synonyms agents emit. Honors DRY_RUN, uses the backend abstraction throughout, naturally idempotent (next tick finds no synonyms). Minor nit-not-blocker: the `pr` / `issue` branches in `_rename_label_on_target` are identical — the case can collapse later. Deferred bats coverage is reasonable given the reconciler is not currently sourceable; tracking under #168 is the right place. CI lint is green. Transitioning to needs-qa.

svv2014:
### 🚫 Judge Verdict: `blocked`

**Points:** +0
**Summary:** Judge could not parse PR outcome.

PR #179, opened 2026-04-30
[LOOP] dep fanout + handler guard + richer dep parsing

svv2014:
Closes the "ticket parked waiting on dep, dep merges, original never wakes" pattern. Three coordinated changes; all keep state in GitHub (no separate queue), inspired by Symphony's "re-poll + reconcile each tick" design but adapted to loop's label-driven model.

## 1. Richer dep parsing — `lib/dep_parser.py` (new)

Extracts blocker refs from issue/PR bodies. Recognises:
- `## Dependencies` section (preserved)
- `blocked by #N` / `blocked-by #N`
- `depends on #N` / `depends-on #N` / `depends #N`
- `requires #N`
- `waiting on #N` / `waits on #N`
- `after #N`

Excludes `closes` / `fixes` / `resolves` (those are scope, not blockers). Self-references filtered.

`recovery_check_dependencies` (lib/recovery.sh) now imports this module — both issue and PR scan paths share one parser instead of two copies of inline regex that only handled the section heading.

## 2. On-merge dep fanout — `scripts/merge-handler.sh`

After PR merge and linked-issue close, calls `recovery_check_dependencies "$SLUG"`. Re-evaluates every blocked ticket immediately (push) instead of waiting up to 15 min for the next reconciler tick (pull).

## 3. Tick-time handler guard — `lib/handler_guard.sh` (new)

Handlers re-fetch live state before doing real work and bail cleanly if the trigger label is gone or the ticket is closed/merged. Catches:
- operator manually relabels between scan and handler dispatch
- another handler closed the ticket in the meantime
- synonym-rename (PR #178) renamed the trigger out from under us

Wired into `dev-handler.sh` and `review-handler.sh`. Other handlers can adopt incrementally.

## 4. Plumbing

`lib/env.sh` exports `LOOP_ROOT` so python subprocesses in recovery.sh can locate `lib/`.

## Test

- `bats tests/dep_parser.bats` — 8/8 (covers each form, exclusions, self-filter, "after that" false-positive guard)
- `bash -n` clean on all modified scripts
- Reconciler `--dry-run --slug loop` exercises the new parser path without error

## Why three at once

The three pieces overlap: the parser feeds both the existing reconciler check and the new on-merge fanout, and the handler guard becomes meaningful only once we trust those two to keep label state coherent. Splitting would mean three PRs that depend on each other landing in sequence — exactly the "park waiting on dep" pattern this PR fixes.


svv2014:
Requesting changes — convention violation that's worth a deliberate decision.

The implementation is sound on its merits: the parser handles the listed forms correctly, exclusions and self-filter behave as specified, the on-merge fanout closes a real "park-on-dep" gap, and the handler guard pattern is clean. Tests cover the regex matrix well (8/8 ✓) and CI lint is green.

But this PR adds `lib/dep_parser.py` as a standalone 67-line Python file, and `CLAUDE.md` is explicit:

> Python snippets are inlined via heredoc (`<<'PY'`) — do not introduce `.py` source files

That rule has been enforced consistently across `lib/`, `scripts/`, and `scanner/` — every Python block in the repo today is a heredoc. Quietly breaking the convention in this PR risks turning it into "the convention until the next time someone wants a real module" without an explicit operator decision.

Two acceptable resolutions:

**Option A (preserve the convention).** Move the parser body into a heredoc inside `lib/recovery.sh` (and a tiny shared helper that both `recovery.sh` and `merge-handler.sh` source). The bats suite can stay — point it at the heredoc-driven helper instead of a `.py` module.

**Option B (change the convention).** If `lib/dep_parser.py` is the right shape because the parser is non-trivial and shared by two call-sites, update `CLAUDE.md` in this PR to reflect that policy change ("Python helpers go in `lib/*.py`; CLI snippets stay heredoc-inlined" or similar), and confirm `install.sh --bootstrap` doesn't need to grow a python-file-specific check. This is a real architectural decision and should be a one-line CLAUDE.md edit, not an implicit slip.

Either path is fine — the issue is that the current diff makes the choice without making the choice. Pick one and re-push. The rest of the PR (env.sh LOOP_ROOT export, on-merge fanout in merge-handler.sh, handler_guard.sh wired into dev/review handlers, recovery.sh deduping the inline regex) is approve-ready.

svv2014:
Addressed review feedback (Option A — preserve the heredoc convention):

- Deleted `lib/dep_parser.py`.
- Added `lib/dep_parser.sh`: parser source lives in a `cat <<'PY' ... PY` heredoc loaded into the shell variable `_DEP_PARSER_PY`. One definition, shared by both scan paths in `recovery.sh`.
- `lib/recovery.sh` now sources `dep_parser.sh` and injects `$_DEP_PARSER_PY` into its issue-scan and PR-scan heredocs via env + `exec()` — `extract()` lands in scope, no `sys.path` munging or module imports.
- Reverted the `LOOP_ROOT` export in `lib/env.sh` — that plumbing existed solely so the python subprocess could find `lib/dep_parser.py`. Not needed anymore.
- `tests/dep_parser.bats` re-pointed at the heredoc-driven helper (sources `dep_parser.sh`, runs a small `parse` helper using the same `DEP_PARSER_PY`/`exec` pattern). All 8 tests pass.

`bash -n` and `shellcheck -S warning` clean on the touched files.

svv2014:
Reviewer (re-review after rework, self-PR cannot use formal approve): LGTM. Rework commit `204e9f1` resolved the convention violation cleanly via Option A — `lib/dep_parser.py` removed, `lib/dep_parser.sh` added as a heredoc-loaded shell helper exporting `_DEP_PARSER_PY`, env.sh LOOP_ROOT export reverted, bats suite re-pointed at the shared helper. 8/8 dep_parser tests pass; `bash -n` clean. The on-merge fanout, handler guard wiring, and recovery.sh dedup that I was already happy with are unchanged. Transitioning to needs-qa.

svv2014:
### 🚫 Judge Verdict: `blocked`

**Points:** +0
**Summary:** Judge could not parse PR outcome.

PR #193, opened 2026-05-01
[LOOP] reconciler: make synonym-rename workflow-aware

svv2014:
Fixes the regression I introduced in #178: `reconcile_synonym_labels` was applying its rename map globally regardless of workflow, silently stripping live trigger labels on any project not on `workflow: default`.

## Observed

svv2014/pa-scanner uses `workflow: current` where `review-pending` IS the PR review trigger. Five PRs (#41, #42, #53, #54, #55) sat starved ~24h while the renamer rewrote `review-pending → needs-review` every tick. Same risk on ppl-study, NanoTraderCopilot, vrefm-classifier.

## Change

`reconcile_synonym_labels` now takes `<slug>` (in addition to `<repo>`) and queries `loop_polled_labels` for the project's actual issue + PR trigger labels per its workflow YAML. For each candidate rename, per-kind:

- skip if `from` IS a trigger label for this project
- skip if `to` is NOT a trigger label for this project

So on a `default`-workflow project, `review-pending → needs-review` still happens. On a `current`-workflow project, it's a no-op.

## Verification

- `bash -n scanner/reconciler.sh` — clean
- The 5 starved pa-scanner PRs were manually relabeled `needs-review → review-pending` to unstick them right now
- Will be exercised live on next reconciler tick after merge

PR #10, opened 2026-04-27
[LOOP-3] projects.yaml v1 schema — workflow ref + label overrides

svv2014:
## Summary

- **`lib/config.sh`**: parse v1 schema fields — `version` (emit deprecation warning when absent), `workflow` ref (default: `"default"`), sparse `labels` map; substitute `${HOME}` in `root`; export `WORKFLOW` and `LOOP_LABEL_OVERRIDES` (pipe-separated `canonical=override` pairs)
- **`scripts/validate-config.sh`**: new validator script — checks slug/repo uniqueness, workflow file existence, label key validity, warns on `${HOME}` placeholders in root paths; wired into `install.sh` as step 2b
- **`config/projects.example.yaml`**: added annotated label-override example block
- **`tests/config.bats`**: 13 unit tests covering v1 default workflow, custom workflow refs, multiple label overrides, v0 legacy fallback, and error cases (all passing)

## Test plan

- [x] `bash -n` passes on all `lib/*.sh scripts/*.sh scanner/*.sh install.sh`
- [x] `shellcheck -S warning` passes on changed files
- [x] `bats tests/config.bats` — 13/13 tests pass
- [ ] Manually run `./scripts/validate-config.sh config/projects.example.yaml` and confirm OK output
- [ ] Confirm `loop_load_project` exports `WORKFLOW` and `LOOP_LABEL_OVERRIDES` correctly for a v1 config

Closes #3

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #204, opened 2026-05-01
[LOOP] fix: _autopull_loop must return 0 — set -e killed reconciler when HEAD unchanged

svv2014:
## Summary

`scanner/reconciler.sh::_autopull_loop` ended with:
```
[ "$before" != "$after" ] && log "auto-pull: updated $before → $after"
```

When `git pull --ff-only` made no change (i.e. nothing to pull), `before == after`, the conditional returns 1, the function inherits exit status 1, and `set -euo pipefail` (top of file) kills the entire script before any reconcile_* check runs. The lock-file EXIT trap fires, then the script exits 1.

## Impact

The reconciler has been silently exiting on every launchd tick for ~2 hours since the most recent merge to main (PR #193, 2026-05-01 01:37). Symptoms:

- `launchctl print gui/$UID/com.user.loop-reconciler` → `last exit code = 1`
- `loop-reconciler.log` last entry at 01:37; nothing for 8+ launchd retries since
- All downstream handlers idle (no events emitted from reconciler-driven paths)
- New project (suprun.ca) added to projects.yaml during the window — never bootstrapped (no labels created on the repo, no events for issue #10)

The bug is deterministic: it triggers on the first reconciler tick after the last merge and stays triggered until something else merges main. Working all day was an artifact of frequent merges; the bug surfaced as soon as merge cadence stopped.

## Fix

Replace the `&&` short-circuit with `if/then` and add explicit `return 0`:

```bash
if [ "$before" != "$after" ]; then
    log "auto-pull: updated $before → $after"
fi
return 0
```

No behavior change in the happy path; only fixes the silent exit when nothing was pulled.

## Verification

- `bash -n scanner/reconciler.sh` clean
- Reproduced locally with `env -i HOME=$HOME PATH=... bash -x reconciler.sh` showing the script exits right after `_autopull_loop` returns when `before == after`
- After fix, manual run + launchd kickstart both proceed past the autopull and run the project loop

## Priority

p0-critical — the entire reconciler-driven pipeline is currently stalled. Recommend admin-merge.

### DIFF
--- a/scanner/reconciler.sh
+++ b/scanner/reconciler.sh
@@ -1545,7 +1545,10 @@ _autopull_loop() {
     before=$(git -C "$LOOP_ROOT" rev-parse HEAD 2>/dev/null || echo "")
     git -C "$LOOP_ROOT" pull --ff-only origin main --quiet 2>/dev/null || return 0
     after=$(git -C "$LOOP_ROOT" rev-parse HEAD 2>/dev/null || echo "")
-    [ "$before" != "$after" ] && log "auto-pull: updated $before → $after"
+    if [ "$before" != "$after" ]; then
+        log "auto-pull: updated $before → $after"
+    fi
+    return 0
 }
 $DRY_RUN || _autopull_loop
 


svv2014:
Self-merging via --admin: this is the bug currently blocking the entire reconciler pipeline (silent exit on every launchd tick for the last 2h). Loop's own automation can't review/merge this because the reconciler is what bootstraps + drives downstream handlers — chicken/egg, same kind of break we hit yesterday with PR #156. Lint passes. Diff is 4 lines.

PR #207, opened 2026-05-01
[LOOP] reconciler: add canonical needs-po/in-po/needs-dev/in-dev to LOOP_PIPELINE_LABELS

svv2014:
## Summary

`LOOP_PIPELINE_LABELS` (scanner/reconciler.sh:581) is the set `reconcile_lost_issues` uses to decide whether an issue carries any pipeline label. It was last updated before the vocabulary migration (#167/#188) introduced `needs-po`/`in-po`/`needs-dev`/`in-dev` as canonical names — those four are missing from the set.

## The ping-pong

After PR #204 brought the reconciler back up:

1. `reconcile_alias_renames` rewrites `po-review → needs-po` (correct for default workflow per #188)
2. Next tick: `reconcile_lost_issues` looks at the issue. Its `LOOP_PIPELINE_LABELS` set doesn't include `needs-po` → false-classifies the issue as "no pipeline label" → routes back to `po-review`
3. Loop, every ~20 min

## Evidence

svv2014/suprun.ca#10 — **15 routing comments over 5 hours** since the reconciler came back up:

```
[2026-05-01T07:51:50Z] Reconciler: issue had no pipeline label — routed back to po-review for triage.
[2026-05-01T08:10:45Z] Reconciler: issue had no pipeline label — routed back to po-review for triage.
... (15× same line, every ~20 min) ...
[2026-05-01T12:35:20Z] Reconciler: issue had no pipeline label — routed back to po-review for triage.
```

Affects every default-workflow project after the alias-rename runs once on a `po-review` or `dev` label.

## Fix

Add `${LOOP_LABEL_NEEDS_PO} ${LOOP_LABEL_IN_PO} ${LOOP_LABEL_NEEDS_DEV} ${LOOP_LABEL_IN_DEV}` to the constant. One-line change, fully additive — no behaviour change for existing labels.

## Verification

- `bash -n scanner/reconciler.sh` clean
- Live impact on next tick: suprun.ca#10's "no pipeline label" log line should disappear; the issue retains `needs-po` and gets dispatched to PO handler

Issue #205, opened 2026-05-01
[bug] reconcile_alias_renames is workflow-blind — strips live triggers on workflow=current projects


svv2014:
## Summary

Same class of bug as #192 (synonym renamer fixed in PR #193), but in the **deprecated-alias renamer** (PR #182's `reconcile_alias_renames` in scanner/reconciler.sh:777). Iterates `LOOP_DEPRECATED_ALIAS_MAP` from lib/labels.sh and applies the rewrites globally without checking each project's workflow.

## Why it bites now

After PR #188 (workflow YAMLs switched to canonical vocab), `config/workflows/default.yaml` uses `needs-po`/`needs-dev`/`needs-review`/`needs-qa` as triggers — aligned with labels.sh canonical. But **`config/workflows/current.yaml` still uses the *old* names** (`po-review`, `dev`, `review-pending`, `changes-requested`, `ready-for-qa`) as triggers.

The renamer treats those names as "deprecated" per labels.sh and rewrites them to canonical — stripping the live trigger labels on every reconciler tick.

## Evidence

```
[2026-05-01 03:46:24] [reconciler] [svv2014/ppl-study] alias-rename issue #423: dev → needs-dev
[2026-05-01 01:34:50] [reconciler] [svv2014/loop] alias-rename issue #199: po-review → needs-po
...
```

For `loop` (workflow=default) the rename is correct — `needs-po` IS the new trigger. For `ppl-study` (workflow=current) the rename is **destructive** — `dev` is the live trigger; `needs-dev` is not.

Affects all 4 projects on `workflow: current`: ppl-study, NanoTraderCopilot, vrefm-classifier, pa-scanner.

## Fix (already in PR)

Mirror PR #193: take slug as second arg, query `loop_polled_labels` for the project's actual triggers per workflow, skip the rename if `from` IS a trigger here OR `to` is NOT a trigger here. Per-kind (issue vs PR).

PR: https://github.com/svv2014/loop/pull/206

## Acceptance

- [ ] Default-workflow projects: `po-review → needs-po` rename still happens (current behaviour preserved)
- [ ] Current-workflow projects: no rewrite of `po-review`/`dev`/`review-pending`/`changes-requested`/`ready-for-qa`
- [ ] Live impact visible on next tick: ppl-study #423's `dev` label stops being renamed
- [ ] Ultimate cure is migrating current.yaml to canonical names too (separate ticket — see #169 / #165 follow-up)

svv2014:
[...]

Issue #199, opened 2026-05-01
[bug] dev-handler / po-handler don't strip issue stage label when opening PR — stale label causes reconciler ping-pong


svv2014:
## Summary

When dev-handler opens a PR closing an issue, it does NOT strip the issue's stage trigger label (`dev`, `po-review`, `in-progress`). The label sits there while the PR works through review/QA. The reconciler then sees an issue with `dev` (or `blocked`, after dev-handler retries) and tries to act on it — re-running dev-handler, or in the recovery path, ping-ponging label transitions.

## Evidence

- **ppl-study #421** (closed yesterday after 7+ recovery ping-pong cycles, 80+ comments): PR #440 was healthy in review the whole time. Issue carried `dev` then `blocked` interchangeably, getting flipped every reconciler tick. Root cause: the issue label should have been stripped when PR #440 opened.
- **loop-monitor #50, #84** today: both stuck with `po-review` while their PRs (#95, #88) are in flight. PO-handler will keep re-running on them.

PR #191 (recovery skip-when-PR-open) is a defensive fix on the consumer side. The structural fix is on the producer side: the handler that opens the PR should mark the issue as "work moved to PR side."

## Fix

In `dev-handler.sh` (and `po-handler.sh` if it ever opens PRs directly), immediately after `gh pr create` succeeds:

```
backend_remove_label "$REPO" "$ISSUE_NUM" "$_DEV_TRIGGER"
backend_remove_label "$REPO" "$ISSUE_NUM" in-progress
# Don't add anything — issue is now in 'work-in-flight-on-PR' state.
# Reconciler should treat 'open issue with no pipeline label AND open closing PR' as healthy.
```

`reconcile_lost_issues` (which routes label-less issues to po-review) must also gain the same gate: skip if the issue has an open closing PR.

## Acceptance

- [ ] After dev-handler opens a PR, the issue carries no stage trigger label (only orthogonal labels: priority, bug, etc.)
- [ ] reconcile_lost_issues skips issues whose closing PR is open
- [ ] reconcile_orphaned_in_progress / reconcile_unblock / reconcile_needs_clarification audited for the same blind spot
- [ ] Bats coverage where the reconciler.sh sourceability allows
- [ ] On merge of this fix: backfill script that strips trigger labels from open issues with open closing PRs (one-shot, dry-run default)

## Related

- #190/#191 (recovery PR-aware) — defensive mirror
[...]

Issue #192, opened 2026-05-01
[bug] reconcile_synonym_labels was workflow-blind — stripped live triggers on workflow=current projects


svv2014:
## Summary

`reconcile_synonym_labels` (introduced in #178) applied its rename map globally without checking each project's workflow. The map renames `review-pending → needs-review`, `ready-for-qa → needs-qa`, `changes-requested → needs-rework`, etc. — correct for projects on `workflow: default`, but **destructive** for projects on `workflow: current`, where `review-pending` / `ready-for-qa` / `changes-requested` ARE the live trigger labels.

## Evidence

svv2014/pa-scanner (`workflow: current`): five PRs (#41, #42, #53, #54, #55) sat starved for ~24h with `needs-review` and zero review-handler activity. Each reconciler tick (~20 min) the renamer rewrote `review-pending → needs-review`. Since `current.yaml` polls `review-pending` for review, the trigger never matched after rename.

Same pattern applies to all other `workflow: current` projects: ppl-study, NanoTraderCopilot, vrefm-classifier.

## Fix (already in PR)

`reconcile_synonym_labels` now takes the slug, queries `loop_polled_labels` for the project's actual trigger labels, and gates each rename:

- skip if `from` IS a trigger for this project (don't strip canonical labels)
- skip if `to` is NOT a trigger for this project (don't apply dead labels)

Per-kind (issue vs PR) since the same label may be a trigger on one side but not the other.

PR: https://github.com/svv2014/loop/pull/193

## Acceptance

- [ ] `default`-workflow project with `review-pending` on a PR → renamed to `needs-review` (current behaviour preserved)
- [ ] `current`-workflow project with `review-pending` on a PR → not touched
- [ ] `current`-workflow project with `needs-review` on a PR → not touched (would only apply if scanner emitted it, but the renamer never adds dead labels)
- [ ] 5 starved pa-scanner PRs (already hand-relabeled to `review-pending`) processed by review-handler on next tick

svv2014:
Reconciler: issue had no pipeline label — routed back to `po-review` for triage.

svv2014:
Closing — fix landed in #193 (merged 2026-05-01 05:21). Auto-close didn't fire because PR body used 'Fixes #192' rather than 'Closes #192' — the merge-handler regex only matches 'closes/closed'. Filing that as a separate bug.

PR #208, opened 2026-05-01
[LOOP] reconciler: consolidate LOOP_PIPELINE_LABELS to lib/labels.sh single source of truth

svv2014:
## Summary

Same vocabulary-migration drift hit us four times today in different files (#192, #205, #207, #199) — each addition during the migration era shipped without checking every other place referencing the same vocab.

Process fix: `lib/labels.sh` now owns the canonical pipeline-tracked-label set. `scanner/reconciler.sh`'s hardcoded `LOOP_PIPELINE_LABELS` string is replaced with a call to a new `loop_pipeline_tracked_labels_string` helper. Future label additions propagate from one place.

## Two distinct sets, both in lib/labels.sh

| Set | Use | Contents |
|---|---|---|
| `LOOP_PIPELINE_STAGE_LABELS` (existing) | strip-on-close | everything in-flight, excludes `blocked`/`done`/`tracker` |
| `LOOP_PIPELINE_TRACKED_LABELS` (new) | "is in the pipeline?" detection | stage + `blocked` + `done` + `tracker` |

## Verification

```
$ source lib/labels.sh
$ loop_pipeline_tracked_labels_string
needs-po in-po needs-dev in-dev needs-review in-review needs-qa qa-pass qa-fail needs-clarification po-review dev plan in-progress review-pending ready-for-qa needs-rework changes-requested in-rework blocked done tracker
```

Same labels as the post-#207 hardcoded string. No behavioural change; refactor only.

## Why

Drift costs went up steeply today: the migration introduced the new canonicals (`needs-po`/`needs-dev`/etc.), but multiple consumers kept their own hardcoded copies of "what counts as a pipeline label". Each one was a separate bug, each one had real impact (suprun.ca#10 ping-ponged 15 times because of one missing entry). One place to edit > four places to remember.

### DIFF
--- a/lib/labels.sh
+++ b/lib/labels.sh
@@ -219,6 +219,31 @@ loop_pipeline_stage_labels_csv() {
     echo "${LOOP_PIPELINE_STAGE_LABELS[*]}"
 }
 
+# Tracked labels — every label that means "this ticket is part of the Loop
+# pipeline at all," including terminal states. Superset of LOOP_PIPELINE_STAGE_LABELS
+# plus blocked/done (terminal) and tracker (epic taxonomy). Used by
+# reconcile_lost_issues to detect issues with NO pipeline label at all (so
+# they can be routed back to po-review for triage).
+#
+# DO NOT use this for strip-on-close — use LOOP_PIPELINE_STAGE_LABELS for
+# that, since blocked/done/tracker must be preserved.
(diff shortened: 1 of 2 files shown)


svv2014:
Self-merging via --admin: refactor that consolidates today's #207 fix into a single source of truth so future vocab additions don't repeat the drift. No behavioural change; manual diff inspection confirms the resolved label set is identical.

PR #213, opened 2026-05-01
[LOOP] reconciler: add canonical needs-po/in-po/needs-dev/in-dev to LOOP_REQUIRED_LABELS

svv2014:
## Summary

Third instance of the same vocabulary-migration drift today. `reconcile_labels` ensures repos have the required Loop labels created as GitHub label objects, but its `LOOP_REQUIRED_LABELS` array never picked up the canonicals introduced in #167/#188.

## Live impact

svv2014/suprun.ca onboarded today → first reconcile_labels run created the old label set without `needs-po`/`needs-dev`/`in-po`/`in-dev`. Every reconciler tick then:

1. alias-rename tries `gh issue edit --add-label needs-po` → fails silently (label doesn't exist on repo)
2. alias-rename `--remove-label po-review` → succeeds
3. Issue ends with NO labels
4. reconcile_lost_issues sees no pipeline label → applies `po-review`
5. Cycle every 20 min — **41+ ping-pong comments on suprun.ca#10**

## Fix

Add the four missing canonical entries to `LOOP_REQUIRED_LABELS`. Next reconcile_labels run on each repo will create them via `gh label create`. Then alias-rename's add succeeds, rename completes atomically, no more "lost" false-routes.

## Verification

- `bash -n` clean
- Same shape as #207's fix to LOOP_PIPELINE_LABELS — additive 4-line change
- Live test: after merge, suprun.ca will get the four labels created on the next reconciler tick; #10 should converge to `needs-po` and stop ping-ponging within 1-2 ticks


svv2014:
Self-merging via --admin: this is the actual root cause of the suprun ping-pong (and any other newly-onboarded repo's). Same trivial drift pattern as #207. 4-line additive change.

Issue #210, opened 2026-05-01
[refactor] Make scanner/reconciler.sh sourceable from tests — top-level main behind a guard


svv2014:
## Summary

`scanner/reconciler.sh` runs `acquire_lock` + main loop at the top level, so sourcing it from a bats test executes the full reconciler. Result: the 14+ `reconcile_*` functions added in 2026-04 cannot be unit-tested. Multiple recently-merged PRs (#178, #191, #193, #206, #207, #208) shipped without coverage on the reconciler functions they touched, with the note "reconciler.sh sourceability needed first."

## Acceptance

- [ ] Wrap the existing top-level execution (lock acquire, autopull, main project loop) in a `main()` function
- [ ] Replace the bottom of the file with:
  ```bash
  if [ "${BASH_SOURCE[0]}" = "$0" ]; then
      main "$@"
  fi
  ```
- [ ] Verify the reconciler still runs identically when invoked normally (`bash scanner/reconciler.sh --slug X` → same output as today)
- [ ] Verify `source scanner/reconciler.sh` from a bats test exposes the `reconcile_*` functions without acquiring locks or hitting the network
- [ ] Add a single sample bats test covering one previously-untestable function (e.g. `reconcile_synonym_labels` with workflow=current → no-op assertion) to demonstrate the new harness works
- [ ] Document the pattern in CONTRIBUTING.md so new reconciler checks ship with bats coverage by default

## Unblocks

Bats coverage promised in: #178, #191, #193, #206, #207, #208 (~6 tickets that left coverage as a follow-up because of this constraint).

svv2014:
Reconciler: issue had no pipeline label — routed back to `po-review` for triage.

svv2014:
Reconciler: issue had no pipeline label — routed back to `po-review` for triage.

svv2014:
Reconciler: issue had no pipeline label — routed back to `po-review` for triage.

svv2014:
Reconciler: issue had no pipeline label — routed back to `po-review` for triage.

svv2014:
Reconciler: issue had no pipeline label — routed back to `po-review` for triage.

svv2014:
Reconciler: issue had no pipeline label — routed back to `po-review` for triage.

svv2014:
Reconciler: issue had no pipeline label — routed back to `po-review` for triage.

svv2014:
Reconciler: issue had no pipeline label — routed back to `po-review` for triage.

[...]

PR #214, opened 2026-05-01
[LOOP] reconcile_lost_issues: observational only — no auto-mutation (root-cause fix for label ping-pong)

svv2014:
## Root cause of today's repeated label ping-pong incidents

Today we patched three separate label-vocabulary drift bugs (#207, #208, #213). All three made the symptoms go away briefly, then a different version of the same ping-pong reappeared. The structural cause was untouched: **`reconcile_lost_issues` is the only reconciler check that mutates state based on a single tick's snapshot, without corroboration**. Every other writer (alias-rename, synonym-rename, recovery_check_*) has gating (workflow triggers, deps satisfied, retry-counter-exhausted, timeout-based). lost-issues has none.

Within one tick, lost-issues runs **after** alias-rename + synonym-rename (which legitimately rewrite labels). Three converging failure modes:

1. **GitHub eventual consistency** — alias-rename's mutation may not yet be visible to lost-issues' subsequent `gh issue list` in the same tick → false-classifies the issue as "no pipeline label."
2. **Stale label-set constants** (today's #207) — needs-po was canonical but `LOOP_PIPELINE_LABELS` didn't include it for weeks → even after a successful rename, lost-issues didn't recognize the result.
3. **Bootstrap drift** (today's #213) — `LOOP_REQUIRED_LABELS` missed needs-po → repo didn't have the label defined → alias-rename's add silently failed → issue ended bare → lost-issues re-routed.

Each subsystem patch made the *current* failure mode go away, but the next eventual-consistency hiccup or future label addition would re-trigger the same class of bug. The fix has to address the writer, not its inputs.

## Change

`reconcile_lost_issues` becomes **observational only**:

- Log "LOST issue #N — observational, no mutation"
- Signal-notify the operator once per (repo, issue), with a 24h cool-down
- Operator applies a trigger label (or closes the issue) — manual decision

No auto-`backend_add_label`. No auto-comment on the issue. No fights with other reconciler writers.

## Behavioural change

Tickets filed without a pipeline label no longer auto-routed to po-review. **This is intentional** — the auto-routing was the bug. The operator-in-the-loop decision was free in practice today (every routing was followed by an operator triage anyway).

Configurable env vars:
- `LOOP_LOST_NOTIFY_HOURS` (default 24)
- `LOOP_LOST_STATE_DIR` (default /tmp/loop-lost-notified)

## Why this matches Symphony's design

Yesterday's discussion: deterministic checks should surface anomalies to operator, not auto-decide. Symphony's reconcile loop is a state machine that reports; humans (or the higher-level claim machine) act. Loop's other checks already follow this; lost-issues was the outlier.

## Verification

- `bash -n scanner/reconciler.sh` clean
- Live test after merge: suprun.ca#10 will receive ONE Signal notification on next tick, then no further mutations until 24h pass or operator labels it
- Bats coverage deferred per #210 (sourceability blocker)

### DIFF
--- a/scanner/reconciler.sh
+++ b/scanner/reconciler.sh
@@ -586,8 +586,23 @@ reconcile_worktrees() {
 LOOP_PIPELINE_LABELS=$(loop_pipeline_tracked_labels_string)
 
 reconcile_lost_issues() {
+    # Observational, NOT mutating. Previously this function applied po-review
+    # to any issue it considered "lost" — but that fought every other writer
+    # within the same tick (alias-rename, synonym-rename, recovery_check_*,
+    # etc.) and could be wrong on stale data (GitHub's read-after-write
+    # consistency window for gh issue edit → gh issue list is non-zero).
+    #
+    # Concrete failure: today on suprun.ca#10 the cycle was alias-rename
+    # writes po-review → needs-po, lost-issues' next gh-fetch in the same
+    # tick missed the update OR labels.sh's pipeline-label set was stale,
+    # lost-issues re-applied po-review, alias-rename re-renamed, ...
+    # 41+ comments, infinite ping-pong.
+    #
+    # Now: log + notify only. The operator (Signal) decides whether the
+    # issue is genuinely abandoned vs. mid-transition. No mutation = no
+    # tick-level coordination needed = no fights with other checks.
     local repo="$1"
-    log "[$repo] scanning for lost issues (no pipeline label)"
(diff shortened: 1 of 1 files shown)


svv2014:
Self-merging via --admin: this is the root-cause fix for the label ping-pong class of bug. Three earlier fixes today (#207, #208, #213) were patches on the inputs; this is the actual writer-side fix that prevents recurrence.

Issue #413, opened 2026-05-15
Scanner can silently stop emitting events — needs liveness heartbeat + auto-restart


svv2014:
## Problem

The scanner process can become silently wedged — alive PID, sleep loop intact, but **emits no events for hours**. Until manually killed, the entire pipeline stops dispatching new work even though all handlers are healthy.

Concrete incident (this session):
- Scanner PID 65773 had been running for **1 day, 5 hours**.
- Last dedup file written to `/tmp/loop-scanner-dedup/` at 05:07:33; observed at 05:45 → 40+ minutes with zero emit activity.
- StandardOutPath (`loop-scanner.log`) was **0 bytes since May 14 21:14** — over a day of no logs.
- `ps` showed process alive, in `sleep 300`. Process tree looked normal.
- Manually killing the PID → launchd auto-restarted (KeepAlive=true); new scanner began logging and emitting events within seconds.

Effect: silent total outage. No alert, no telemetry, no error. The only way to know is to compare `loop-scanner.log` mtime against `now`.

## Hypothesis

Three plausible failure modes (not yet pinpointed):

1. Sleep child returns but parent's main loop body is wedged on an inner subprocess (e.g. a `gh` call that timed out via TCP layer, not handled by handler-level timeouts).
2. STDOUT got redirected to a closed pipe upstream (a dead `tee` child), so subsequent `echo`s in `log()` block on EPIPE in a way that hangs the script.
3. A previous `set -e` triggered exit-via-trap that didn't propagate; the loop continued but with a broken state.

## Acceptance

Scanner must be self-monitoring. Specifically:

1. **Heartbeat file** — every tick, write `now` to `${LOOP_LOG_DIR}/scanner-heartbeat`. A separate watchdog (launchd or cron) reads the file; if `now - mtime > 2 * poll_interval`, kills the scanner PID and restarts it.
2. **Tick-counter event** — every tick emit a `scanner_tick` event with `now` + `dedup_count` to loop-monitor. Monitor surfaces it on the dashboard so silent-stop is visible.
3. **Stdout file integrity check** — at the top of each tick, if `[ ! -w "$LOG_FILE" ]` or `[ ! -t $STDOUT_FD ]`, recover the file descriptor or exit so launchd restarts.

[...]

PR #400, opened 2026-05-15
[LOOP-396] require Closes #N on every dev-handler PR + CI gate

svv2014:
Closes #396

### `scripts/dev-handler.sh`
Added a post-PR-open validation step in the belt-and-braces success path (after the agent has run and a PR has been found):

1. Fetches the PR body via `backend_pr_view`.
2. Applies regex `(?im)^(closes|fixes|resolves)\s+#([0-9]+)\b` to extract the closing issue number.
3. If no reference is found — closes the PR (with branch delete), posts an explanatory comment on the originating issue, emits `dev_failed_no_ticket` via `bounty_report`, and labels the issue `needs-clarification`.
4. If a reference is found but the referenced issue is not open (`gh issue view ... --json state`) — same abort flow with a different error message.
5. Only if both checks pass does the handler proceed to the existing label-swap logic.

### `.github/workflows/check-closes-ref.yml` (new)
CI gate — defence in depth. Runs on every non-bot `pull_request` event (opened, synchronize, reopened, edited):

- Skips PRs authored by known bot actors (`github-actions[bot]`, `dependabot[bot]`, `loop-bot`).
- Skips PRs carrying the `release-pr` label.
- Extracts the first `Closes/Fixes/Resolves #N` reference from the PR body.
- Fails with a descriptive `::error::` message if none is found.
- Validates the referenced issue is open via `gh api`; fails if it is closed or not found.

## Test Plan

- Open a PR without any `Closes #N` line → CI `check-closes-ref` job must fail.
- Open a PR with `Closes #999999` (non-existent issue) → CI job must fail.
- Open a PR with `Closes #396` (open issue) → CI job must pass.
- Open a PR authored by `github-actions[bot]` or labelled `release-pr` → CI job skips.
- Trigger dev-handler on an issue; confirm handler closes any agent-opened PR that lacks the reference and posts the expected comment.

Issue #396, opened 2026-05-15
Dev handler must require Closes #N on every opened PR (with CI gate)


svv2014:
## Problem

Loop's dev handler will open and merge PRs that have no `Closes #N` line, so work lands without a ticket of record. Sample from loop-monitor's last 30 merges: ~15 PRs have no `Closes`. Examples:

- `#304 feat(dashboard): summary + activity layout, fits 80×24` — no ticket
- `#305 feat(dashboard): active workers, stage strip, 24h sparkline, stuck list` — no ticket, merged 5 minutes after #304 on the same file
- `#262 fix(web): set vite base to / after Phase 5 cutover (#259 follow-up)` — references a PR, not an issue

Effect: no audit trail for *why* the change shipped, the pipeline's "Closes #N → close issue on merge" wiring silently does nothing, and operator priorities can't gate work that bypasses the queue.

## Acceptance

Dev handler refuses to open / push a PR whose body lacks a parseable `Closes #N` (or `Fixes #N`, `Resolves #N`) line that resolves to an open issue in the **same** repo.

- Body parser: regex `(?im)^(closes|fixes|resolves)\s+#(\d+)\b` — first capture group.
- Validate the issue exists and is open via `gh issue view <N> --json state`.
- If missing or unresolvable: handler aborts the PR open, posts a comment on the issue (if it was triggered from one), and emits a `dev_failed_no_ticket` event.

CI gate (defence in depth) — add a workflow that fails any non-bot PR opened to the default branch with the same check.

## Out of scope

- Retroactively backfilling tickets for already-merged PRs.
- Allowing `Closes #N` on a different repo (cross-repo work is rare; defer).
- Bot-authored release PRs (whitelist via author or label `release-pr`).

PR #416, opened 2026-05-15
fix(dev-handler): normalize gh issue state case (uppercase OPEN broke #396 check)

svv2014:
Related to #413 (triggered by the uppercase-OPEN bug, but fixes a separate concern)

## Problem

Every dev PR opened since #400 (LOOP-396) merged ~1h ago has been auto-closed by dev-handler with the error "`Closes` references a non-open issue" — even when the referenced issue was open. The rejection comment on PR #414 (for #413) literally read:

> Closes #413 but that issue is not open (state: `OPEN`).

`gh issue view --json state --jq .state` returns `"OPEN"` / `"CLOSED"` (uppercase). The dev-handler check at line 335 compared against `"open"` (lowercase), so the check failed for every actually-open issue.

## Fix

Normalize to lowercase before comparing. One-line change.

## Test plan
- `bash -n scripts/dev-handler.sh` → passes.
- Manual verify: `printf '%s' OPEN | tr '[:upper:]' '[:lower:]'` returns `open`; comparison passes.
- Follow-up: dev-handler should re-process #413 successfully once this lands.

### DIFF
--- a/scripts/dev-handler.sh
+++ b/scripts/dev-handler.sh
@@ -332,7 +332,9 @@ print(m.group(2) if m else '')
         fi
 
         _closes_state=$(gh issue view "$_closes_num" --repo "$REPO" --json state --jq .state 2>/dev/null || echo "")
-        if [ "$_closes_state" != "open" ]; then
+        # gh returns "OPEN"/"CLOSED" (uppercase); normalize before comparing.
+        _closes_state_lc=$(printf '%s' "$_closes_state" | tr '[:upper:]' '[:lower:]')
+        if [ "$_closes_state_lc" != "open" ]; then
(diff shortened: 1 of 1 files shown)


svv2014 (commented):
[loop:review] The one-line fix is correct — `gh issue view --json state` does return uppercase `"OPEN"`/`"CLOSED"`, and normalizing to lowercase before the comparison is the right minimal approach. CI passes (lint + check-closes-ref).

However, `Closes #413` in the PR body is wrong: issue #413 tracks the scanner watchdog/heartbeat feature (heartbeat file, tick-counter events, stdout integrity check), none of which are implemented here. If this merges as-is, GitHub will auto-close #413, marking that unimplemented feature as done.

**Required change:** Remove `Closes #413` from the PR body. Replace with a note like "Unblocks re-processing of #413" so the scanner watchdog issue stays open for a follow-up PR.

PR #424, opened 2026-05-15

svv2014:
Closes #413

## Changes

**`scanner/scanner.sh`** — `run_once()` now calls `touch ${LOOP_LOG_DIR}/scanner-heartbeat` at the start of every tick (skipped in `--dry-run` mode). This creates a timestamp file that external watchers can use to detect wedged scanners.

**`scripts/scanner-watchdog.sh`** — New script. On each run it reads the heartbeat file mtime, computes age against `2 × LOOP_SCANNER_INTERVAL` (default 600s). If stale: reads the scanner PID from `/tmp/loop-scanner.lock` and kills it so launchd KeepAlive (or the next cron invocation) restarts the scanner. Supports `--dry-run` for safe testing.

**`templates/launchd/com.user.loop-scanner-watchdog.plist.template`** — New launchd plist template. Runs watchdog every 5 min via `StartInterval=300`.

**`install.sh`** — `bootstrap_register_launchd()` installs the watchdog plist; `bootstrap_register_cron()` adds a `*/5` cron entry. Both are idempotent (skip if already present).

**`tests/scanner-heartbeat.bats`** — 6 BATS tests:
- `run_once` writes heartbeat file
- `run_once` refreshes heartbeat mtime on repeated calls
- `run_once` skips heartbeat in `--dry-run` mode
- watchdog exits 0 when heartbeat is fresh
- watchdog `--dry-run` reports stale without killing
- watchdog exits 0 when no heartbeat file exists yet

## Test Plan

- All 6 BATS tests pass locally
- `bash -n` and `shellcheck -S warning` clean on all scripts
- No personal identifiers detected by CI grep
- Manual: simulate wedge with `kill -STOP $SCANNER_PID` → watchdog kills it within 10 min; launchd restarts within 60s

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -338,6 +338,25 @@ bootstrap_register_launchd() {
         fi
     fi
 
+    # Scanner liveness watchdog — kills a wedged scanner so launchd restarts it.
+    # Runs every 5 min; checks heartbeat age against 2 × LOOP_SCANNER_INTERVAL.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #429, opened 2026-05-15
[LOOP-413] scanner liveness heartbeat + auto-restart watchdog

svv2014:
Closes #413

### scanner/scanner.sh
- `run_once()` now writes the current timestamp to the heartbeat file at the start of every tick (before any work), so the watchdog can detect wedges even when the tick completes but the log is silent
- Added stdout integrity check: if `LOG_FILE` is not writable at tick start (dead tee child / broken pipe), the script reopens FDs 1+2 against the log path; exits if reopen fails so launchd can restart cleanly
- Logs dedup entry count at tick end (`dedup_entries=N`) as the tick-counter signal

- Reads heartbeat file mtime; if age > `LOOP_SCANNER_WATCHDOG_THRESHOLD` (default 900s / 15 min), the scanner is considered wedged
- macOS: kills the scanner PID — launchd KeepAlive triggers automatic restart
- Linux: kills the PID and relaunches `scanner.sh` in the background directly (no launchd)
- Removes stale lock file if PID is already gone so the next invocation can acquire the lock
- `--dry-run` flag for safe testing

- Runs `scanner-watchdog.sh` every 5 minutes via `StartInterval`

### install.sh
- `bootstrap_register_launchd()`: registers `com.user.loop-scanner-watchdog` plist (idempotent)
- `bootstrap_register_cron()`: adds `*/5 * * * *` watchdog cron entry (idempotent, Linux path)

## Test Plan
- Manual: run `scanner.sh`, verify `scanner-heartbeat` file appears in `$LOOP_LOG_DIR` after first tick
- Manual simulate wedge: `kill -STOP $SCANNER_PID`, wait > 15 min (or lower threshold via `LOOP_SCANNER_WATCHDOG_THRESHOLD`), run `scanner-watchdog.sh` → scanner restarts
- `scanner-watchdog.sh --dry-run` when scanner is healthy → exits 0, logs "scanner is live"
- `scanner-watchdog.sh --dry-run` with removed heartbeat → logs stale warning, does not kill
- Verify `bash -n` and `shellcheck` pass on all modified scripts

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,25 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — kills a wedged scanner every 5 min so the
+    # KeepAlive scanner plist restarts it automatically (#413).
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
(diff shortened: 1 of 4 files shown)


svv2014:

PR #431, opened 2026-05-15
[LOOP-413] add scanner liveness heartbeat and auto-restart watchdog

svv2014:
Closes #413

## Changes

**scanner/scanner.sh**
- Added `HEARTBEAT_FILE` written via `touch` at the start of every `run_once()` tick (skipped in --dry-run mode).
- Added log-FD integrity check at the top of each tick: if LOG_FILE is no longer writable, reopens FDs 1+2; if that fails, exits so launchd/cron can restart cleanly.

- Reads heartbeat file mtime; if absent or older than LOOP_SCANNER_WATCHDOG_STALE seconds (default: 2 * LOOP_SCANNER_INTERVAL = 600s), kills scanner PID from the lock file.
- On macOS: calls launchctl kickstart -k for immediate restart.
- On Linux: removes stale lock; cron restarts scanner on next 5-min tick.
- Supports --dry-run flag.

- Runs scanner-watchdog.sh every 5 minutes (StartInterval 300).

**install.sh**
- bootstrap_register_launchd: registers com.user.loop-scanner-watchdog.plist.
- bootstrap_register_cron: adds watchdog cron entry (*/5 min) for Linux.

**loop.env.example**
- Documents LOOP_SCANNER_WATCHDOG_STALE (commented out; default is 2x poll interval).

## Test Plan

- bash -n and shellcheck pass on all scripts
- Simulate wedge: kill -STOP $SCANNER_PID -> heartbeat goes stale -> watchdog kills and restarts within one 5-min tick
- Verify scanner-heartbeat file is updated on every scanner tick
- scanner-watchdog.sh --dry-run prints intended actions without killing anything
- ./install.sh --bootstrap registers the watchdog plist on macOS and cron entry on Linux (idempotent)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,25 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — checks heartbeat every 5 min and restarts the
+    # scanner if it has been silent for more than 2 * LOOP_SCANNER_INTERVAL seconds.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #435, opened 2026-05-15
[LOOP-413] add scanner liveness heartbeat and watchdog

svv2014:
Closes #413

## Changes

**scanner/scanner.sh**
- At the start of every `run_once()` tick, write the current epoch to `${LOOP_LOG_DIR}/scanner-heartbeat` (skipped in dry-run mode). This gives the watchdog a reliable signal that the scanner is actively polling.
- Added stdout integrity check: if `LOG_FILE` exists but is no longer writable (e.g. after a log rotation that orphaned the FD), attempt to reopen it; exit if reopen fails so launchd auto-restarts the process.

- Runs every 5 min. Reads `scanner-heartbeat` mtime; if age > `2 × LOOP_SCANNER_INTERVAL` (default 10 min), kills the scanner PID, removes the stale lock, and nudges `launchctl kickstart` on macOS. Supports `--dry-run`.
- On Linux (cron mode), killing the PID is a no-op since cron handles scheduling — the watchdog still cleans the stale lock if present.

- launchd `StartInterval=300` plist for the watchdog (macOS).

**install.sh**
- `bootstrap_register_launchd`: registers `com.user.loop-scanner-watchdog` alongside the existing scanner/reconciler/pr-watchdog/digest agents.
- `bootstrap_register_cron`: adds a `*/5` crontab entry for `scanner-watchdog.sh` on Linux.

**tests/scanner-heartbeat.bats** (new)
- 7 tests: heartbeat file is created on first tick, contains a recent epoch, is refreshed on repeated ticks, is absent in dry-run mode, watchdog reports OK for a fresh heartbeat, reports stale for an old one, exits cleanly when the file is missing.

## Test Plan
- CI: `bash -n`, `shellcheck`, personal-identifier grep — all pass locally.
- bats `tests/scanner-heartbeat.bats` — 7/7 pass.
- Manual: `kill -STOP $SCANNER_PID` → watchdog should restart within 10 min (2×300s).
- Existing `tests/scanner.bats` and `tests/scanner-log-sighup.bats` — no regressions (same 5 pre-existing failures, no new ones).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,25 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — kills and restarts a silently-wedged scanner.
+    # Runs every 5 min; checks heartbeat file mtime against 2×poll_interval.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #436, opened 2026-05-15

svv2014:
Closes #413

## Changes

**`scanner/scanner.sh`** — `_write_heartbeat()` helper writes `date +%s` to `${LOOP_LOG_DIR}/scanner-heartbeat` at the top of every `run_once()` call. Skipped in `--dry-run` mode. The file mtime is the liveness signal: watchdog checks it, monitoring can stat it.

**`scanner/scanner-watchdog.sh`** — new script (runs every 5 min via launchd/cron). Reads the heartbeat mtime; if age ≥ `LOOP_WATCHDOG_STALE_SECONDS` (default 900 s = 15 min, covering 3× the default 5-min poll interval), reads the scanner PID from `/tmp/loop-scanner.lock`, sends SIGTERM (then SIGKILL after 3 s if still alive), and exits. launchd `KeepAlive=true` on the scanner restarts it automatically within seconds.

**`templates/launchd/com.user.loop-scanner-watchdog.plist.template`** — launchd plist (`StartInterval 300`) for macOS deployments. Uses the same `__LOOP_ROOT__`/`__LOG_DIR__`/`__HOME__`/`__EXTRA_PATH__` placeholders as the other templates.

**`install.sh`** — registers the watchdog on both macOS (launchd, idempotent) and Linux (cron `*/5 * * * *`, idempotent marker `# loop-scanner-watchdog`).

**`tests/scanner-heartbeat.bats`** — 8 bats tests covering: heartbeat file creation, timestamp content, mtime update on each call, dry-run no-op, and watchdog behaviour for fresh / stale / no-lock-file cases.

## Test Plan
- All 8 new bats tests pass locally
- Manual simulation: `kill -STOP $SCANNER_PID` → watchdog fires within 5 min, kills the stopped PID, launchd restarts scanner

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,25 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — detects a silently-stalled scanner and kills
+    # it so launchd restarts it automatically. Runs every 5 min independently.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #438, opened 2026-05-16

svv2014:
Closes #413

## Changes

**`scanner/scanner.sh`**
- Added `HEARTBEAT_FILE="${LOOP_LOG_DIR}/scanner-heartbeat"` variable.
- `run_once()` writes `date +%s` to the heartbeat file at the very top of each tick, before any `gh` I/O. A scan that wedges partway through still records its heartbeat, so the watchdog detects the wedge on the *next* interval (≤10 min with defaults).

**`scripts/restart-scanner-if-stale.sh`** (new)
- Watchdog that reads the heartbeat file mtime and compares to `now`.
- Stale threshold: `LOOP_SCANNER_STALE_THRESHOLD` (defaults to `2 × LOOP_SCANNER_INTERVAL` = 600s).
- On stale: reads the PID from `/tmp/loop-scanner.lock`, sends `SIGTERM`, removes the lock so launchd can restart clean.
- On macOS with `KeepAlive=true` the scanner restarts in seconds. On Linux with cron the next `--once` invocation proceeds normally.
- `--dry-run` flag supported for safe manual inspection.

- launchd plist running the watchdog every 5 min (`StartInterval 300`).
- Logs to `${LOG_DIR}/loop-scanner-watchdog.log`.

**`install.sh`**
- `bootstrap_register_launchd`: registers `com.user.loop-scanner-watchdog` between the scanner and reconciler plists.
- `bootstrap_register_cron`: adds a `*/5 * * * *` crontab entry for the watchdog on Linux.

- 7 tests: heartbeat variable presence, file creation on `run_once()`, timestamp format, mtime refresh on repeated calls, watchdog OK/stale/missing scenarios via `--dry-run`.

## Test Plan

- `bash -n` + `shellcheck -S warning` both pass with zero warnings.
- `bats tests/scanner-heartbeat.bats` — all 7 tests pass.
- Smoke: `./scripts/restart-scanner-if-stale.sh --dry-run` while scanner is running → logs `OK: heartbeat age=...`
- Simulate wedge: `kill -STOP $SCANNER_PID` → watchdog detects stale within ≤10 min and kills it; launchd restarts automatically.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -263,6 +263,7 @@ bootstrap_register_launchd() {
     mkdir -p "$agents_dir"
 
     local scanner_plist="$agents_dir/com.user.loop-scanner.plist"
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
     local reconciler_plist="$agents_dir/com.user.loop-reconciler.plist"
     local reconcile_plist="$agents_dir/com.user.loop-reconcile.plist"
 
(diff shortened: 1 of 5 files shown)


svv2014:

PR #450, opened 2026-05-16

svv2014:
Closes #413

### scanner/scanner.sh
- Added `_scanner_write_heartbeat()` — writes the current timestamp to `${LOOP_LOG_DIR}/scanner-heartbeat` on every tick. The watchdog reads this file to detect a wedged scanner.
- Added `_scanner_check_log_writable()` — at the top of each tick, re-opens stdout/stderr against the log file path (defensive against closed/orphaned FDs). If the log directory is entirely unwritable, exits immediately so launchd restarts cleanly.
- Both functions are called at the start of `run_once()` before the scan body.
- `_scanner_write_heartbeat` is a no-op in `--dry-run` mode.

- Fires every 5 min (via launchd or cron). Reads `scanner-heartbeat` mtime; if older than `LOOP_SCANNER_WATCHDOG_STALE` (default 900s = 15 min), kills the wedged scanner PID and triggers a restart.
- macOS: uses `launchctl kickstart` so launchd's KeepAlive immediately respawns the scanner.
- Linux: removes the stale lock file so the next cron invocation starts cleanly.
- Supports `--dry-run` flag; logs to `loop-scanner-watchdog.log`.

- launchd plist for the watchdog agent. `StartInterval=300` (every 5 min).

### install.sh
- `bootstrap_register_launchd`: registers `com.user.loop-scanner-watchdog` alongside existing agents.
- `bootstrap_register_cron`: adds a `*/5 * * * *` cron entry for the watchdog on Linux.

- 5 bats tests: heartbeat file creation, timestamp format, mtime update on repeated calls, dry-run no-op, and `run_once()` integration.

## Test Plan
- `bats tests/scanner-heartbeat.bats` — all 5 pass locally.
- Confirm `${LOOP_LOG_DIR}/scanner-heartbeat` is updated every poll interval after deploy.
- Manual wedge test: `kill -STOP $SCANNER_PID` → watchdog should restart within 15 min (next 5-min tick that sees a 15-min-stale heartbeat).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,25 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — fires every 5 min, restarts scanner if heartbeat
+    # file is older than LOOP_SCANNER_WATCHDOG_STALE (default 15 min).
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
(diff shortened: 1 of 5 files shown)


svv2014:

Issue #194, opened 2026-05-01
[bug] Scanner stdout/stderr log gets orphaned — process writes to deleted inode, on-disk file is 0 bytes


svv2014:
## Summary

`loop-scanner.log` on disk is empty (0 bytes), yet the running scanner process holds an open FD to the same path with ~7MB written. Some external action (logrotate, manual truncation, plist reload) created a new inode at the same path; the scanner kept writing to the original inode, which is now an orphaned hidden file.

Effect: operators investigating pipeline issues have no visibility into scanner decisions (which projects scanned, which events emitted, which dedups skipped). Today this turned a 5-min diagnosis into 30 min for the pa-scanner starvation incident.

## Repro

```
$ ls -la ~/.openclaw/workspace/logs/loop/loop-scanner.log
-rw-r--r-- 1 vadim staff 0 Apr 30 22:02 .../loop-scanner.log

$ ps -ef | grep scanner.sh
501 93785 ... bash .../scanner/scanner.sh

$ lsof -p 93785 | grep log
bash 93785 vadim 1u REG 1,18 7324560 312655142 .../loop-scanner.log
bash 93785 vadim 2u REG 1,18 7324560 312655142 .../loop-scanner.log
```

The lsof inode (312655142) and the disk inode no longer match → orphan.

## Fix options

1. **SIGHUP reopen** (preferred): scanner traps SIGHUP and reopens its log fds (`exec >>"$LOG_FILE" 2>&1`). Logrotate config sends SIGHUP after rotation. Standard daemon pattern.
2. **Append via helper** (next best): every log line goes through a function that opens-appends-closes (`echo "$line" >> "$LOG_FILE"`). Slightly more syscalls, but rotation-proof.
3. **Drop the launchd StandardOutPath redirection and use a stdlib logger** (more invasive).

Option 1 is consistent with how the gateway and other long-lived loop daemons should behave anyway.

## Acceptance

- [ ] Scanner reopens its log file on SIGHUP without restarting
- [ ] Manual `: > loop-scanner.log` followed by `kill -HUP <pid>` results in fresh writes appearing in the file
- [ ] Same handling applied to reconciler.sh + any other long-lived loop daemon writing through launchd StandardOutPath

svv2014:
Reconciler: issue had no pipeline label — routed back to `po-review` for triage.

svv2014:
Reconciler: issue had no pipeline label — routed back to `po-review` for triage.

svv2014:
[...]

PR #454, opened 2026-05-16

svv2014:
Closes #413

### `scanner/scanner.sh`
- Added `HEARTBEAT_FILE` variable pointing to `${LOOP_LOG_DIR}/scanner-heartbeat`.
- At the top of every `run_once()` tick (skipped in `--dry-run` mode), writes the current timestamp to the heartbeat file. This gives the watchdog a reliable signal that the scanner is still executing tick bodies, not just sleeping.
- Added a stdout writability guard at the top of each tick: if `LOG_FILE` is not writable (e.g. inode replaced after log rotation without a SIGHUP), attempts to reopen FDs 1+2 against the path; exits so launchd can restart if reopen fails. Complements the existing SIGHUP reopen trap (#194).

- Reads the heartbeat file's mtime. If older than `LOOP_WATCHDOG_THRESHOLD` seconds (default `2 × LOOP_SCANNER_INTERVAL` = 600s), kills the scanner PID recorded in `/tmp/loop-scanner.lock` (SIGTERM → SIGKILL if needed). launchd's `KeepAlive=true` then restarts the scanner automatically; on Linux, the next cron tick does the same.
- Supports `--dry-run` to log what would happen without killing anything.
- Gracefully skips if the heartbeat file doesn't exist yet (scanner never ran).

- Launchd plist that fires the watchdog every 5 minutes (`StartInterval=300`).
- Uses the same placeholder convention (`__LOOP_ROOT__`, `__LOG_DIR__`, `__HOME__`, `__EXTRA_PATH__`) as the existing scanner plist template.

- 7 tests covering: heartbeat file created on first tick, contains a timestamp, mtime updated on each tick, not written in dry-run mode, watchdog exits healthy on fresh heartbeat, watchdog detects stale heartbeat in dry-run, watchdog skips gracefully with no heartbeat file.

## Test Plan
- All 7 new bats tests pass.
- All pre-existing scanner tests pass (5 pre-existing failures in `scanner.bats` are unrelated to this PR and present on `main`).
- No personal identifiers found.
- Manual: simulate wedge with `kill -STOP $SCANNER_PID` → after `LOOP_WATCHDOG_THRESHOLD` seconds the watchdog should log "wedged" and kill it; launchd should restart within 60s.

### DIFF
--- a/scanner/scanner-watchdog.sh
+++ b/scanner/scanner-watchdog.sh
@@ -1,0 +1,107 @@
+#!/usr/bin/env bash
+# scanner-watchdog.sh — detect a wedged scanner and restart it.
+#
+# The scanner writes a heartbeat timestamp to ${LOOP_LOG_DIR}/scanner-heartbeat
+# on every tick. This watchdog checks that file's mtime; if it is older than
+# STALE_THRESHOLD seconds (default 2× POLL_INTERVAL = 600s) the scanner is
+# considered wedged, its PID is killed, and launchd/cron restarts it.
+#
+# Designed to run every 5 minutes via launchd (macOS) or cron (Linux).
+#
+# Usage:
+#   scanner-watchdog.sh [--dry-run]
+#
+# Environment:
+#   LOOP_SCANNER_INTERVAL   poll cadence in seconds (default 300)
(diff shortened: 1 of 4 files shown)


svv2014:

PR #456, opened 2026-05-16

svv2014:
Closes #413

### `scanner/scanner.sh`
- At the top of each `run_once()` tick, writes the current Unix timestamp to `${LOOP_LOG_DIR}/scanner-heartbeat`. Skipped in `--dry-run` and `--once` modes to avoid side-effects.
- Adds a stdout integrity check: if `LOG_FILE` is non-writable at tick start, logs a warning to stderr and exits so launchd can restart with a fresh fd. This catches the closed-pipe failure mode described in the issue.

### `scanner/watchdog.sh` (new)
- Fires every 5 minutes (via launchd `StartInterval` or cron `*/5`).
- Reads the heartbeat file mtime. If older than `LOOP_SCANNER_STALE_THRESHOLD` seconds (default 900 = 15 min), reads the scanner PID from the lock file and kills it.
- On macOS with `KeepAlive=true`, launchd auto-restarts the scanner within seconds. On Linux, clearing the lock lets the next cron tick start fresh.
- Supports `--dry-run` for testing without actually killing the scanner.

- LaunchAgent plist template; `StartInterval=300` (every 5 minutes).

### `install.sh`
- `bootstrap_register_launchd`: installs and loads the watchdog plist alongside the scanner and reconciler plists (idempotent).
- `bootstrap_register_cron`: adds a `*/5 * * * *` cron entry for the watchdog on Linux (idempotent, marker-guarded).

7 tests covering:
- Heartbeat file created on first tick
- Heartbeat file contains a valid, recent Unix timestamp
- Heartbeat file updated on subsequent ticks
- Heartbeat NOT written in dry-run mode
- `watchdog.sh` passes `bash -n`
- Watchdog `--dry-run` reports "alive" when heartbeat is fresh
- Watchdog `--dry-run` reports "wedged" and would-kill message when heartbeat is stale

## Test Plan
- All 7 new bats tests pass locally
- No personal identifiers detected
- Manual verification: `kill -STOP $SCANNER_PID` → watchdog kills it within 15 min; launchd restarts automatically

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -338,6 +338,26 @@ bootstrap_register_launchd() {
         fi
     fi
 
+    # Scanner liveness watchdog — fires every 5 min; kills a wedged scanner
+    # PID so launchd can auto-restart it (KeepAlive). Complements SIGHUP log
+    # reopen (#194) by catching the case where the process is alive but silent.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #458, opened 2026-05-16
[LOOP-413] scanner: heartbeat file + watchdog auto-restart on silent stop

svv2014:
Closes #413

## Changes

**scanner/scanner.sh** — write current epoch to `${LOOP_LOG_DIR}/scanner-heartbeat` at the start of every tick (skipped in `--dry-run` mode). If the scanner wedges between ticks, this file's mtime stops advancing, which the watchdog uses as the liveness signal.

**scanner/scanner-watchdog.sh** — new script that reads `scanner-heartbeat` mtime and restarts the scanner when it is older than `2 × LOOP_SCANNER_INTERVAL` (default 600 s). On macOS it calls `launchctl kickstart -k` so launchd kills the stuck process and relaunches under KeepAlive; on Linux / fallback it sends SIGTERM to the PID from the lock file. Supports `--dry-run`.

**templates/launchd/com.user.loop-scanner-watchdog.plist.template** — launchd plist that fires the watchdog every 5 min (`StartInterval=300`). Uses the same `__LOOP_ROOT__` / `__LOG_DIR__` / `__HOME__` / `__EXTRA_PATH__` placeholders as the other loop plist templates.

**install.sh** — registers the watchdog plist in `bootstrap_register_launchd` (macOS) and adds a `*/5 * * * *` cron entry in `bootstrap_register_cron` (Linux), both idempotent.

## Test Plan

- `tests/scanner-heartbeat.bats` (4 new tests):
  - heartbeat file is created on first tick
  - file contains a valid 10-digit unix timestamp
  - mtime advances on consecutive ticks
  - file is NOT written in `--dry-run` mode
- CI: `bash -n` + `shellcheck -S warning` pass on all changed scripts
- Manual: simulate wedge with `kill -STOP $SCANNER_PID` → watchdog should restart within 10 min (2× 300 s poll interval)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -316,6 +316,25 @@ bootstrap_register_launchd() {
         fi
     fi
 
+    # Scanner liveness watchdog — restarts scanner if heartbeat goes stale.
+    # Runs every 5 min; only kicks in if heartbeat age > 2× poll interval.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #465, opened 2026-05-16
[LOOP-413] scanner: liveness heartbeat + scanner-watchdog auto-restart

svv2014:
Closes #413

### scanner/scanner.sh
- Added a heartbeat write at the top of `run_once()`: every tick writes the current timestamp to `${LOOP_LOG_DIR}/scanner-heartbeat`. Skipped in `--dry-run` mode. This gives a reliable mtime-based liveness signal with no new dependencies.

- Standalone script that reads the heartbeat file's mtime and compares it against a configurable stale threshold (default: `2 × LOOP_SCANNER_INTERVAL + 60s` = 660s).
- If stale: reads the scanner PID from `/tmp/loop-scanner.lock`, sends SIGTERM, waits 5s, then SIGKILL if still alive, and removes the lock file so launchd/cron restarts a fresh scanner automatically.
- If the heartbeat file doesn't exist yet (scanner hasn't ticked once), age=0 which is always below threshold — considered healthy.
- Accepts `--dry-run` flag (reports but does not kill).
- Threshold tunable via `LOOP_SCANNER_WATCHDOG_STALE` env var.

- Runs `scanner-watchdog.sh` every 5 minutes via launchd (macOS).
- Uses the same `__LOOP_ROOT__` / `__LOG_DIR__` / `__HOME__` / `__EXTRA_PATH__` substitution pattern as the other plist templates.

### install.sh
- `bootstrap_register_launchd()`: registers `com.user.loop-scanner-watchdog` after digest, using the same sed/launchctl pattern as existing agents.
- `bootstrap_register_cron()`: adds a `*/5 * * * *` cron entry for `scanner-watchdog.sh` on Linux, with a `# loop-scanner-watchdog` marker for idempotent re-runs.

7 bats tests covering:
- `run_once` writes heartbeat file
- Heartbeat contains a timestamp in `YYYY-MM-DD HH:MM:SS` format
- Heartbeat mtime is updated on each call
- Heartbeat is NOT written in dry-run mode
- Watchdog `--dry-run` reports healthy when heartbeat is fresh
- Watchdog `--dry-run` reports stale + would-kill when heartbeat is old
- Watchdog `--dry-run` exits 0 when heartbeat file doesn't exist yet (age=0)

## Test Plan

- All 7 new bats tests pass locally.
- All pre-existing scanner tests unchanged (pre-existing failures on `main` are unaffected).
- `bash -n` and `shellcheck -S warning` clean on all scripts.
- No personal identifiers in any changed file.
- Manual simulation: `kill -STOP $SCANNER_PID` for >660s, then check that watchdog kills it and launchd restarts within the next 5-min watchdog cycle.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,25 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — checks heartbeat file every 5 min; kills
+    # and lets launchd restart the scanner if it has been silent too long.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
+            -e "s|__HOME__|$HOME|g" \
+            -e "s|__EXTRA_PATH__|$extra_path|g" \
+            "$template_dir/com.user.loop-scanner-watchdog.plist.template" > "$scanner_watchdog_plist"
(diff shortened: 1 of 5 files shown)


svv2014:

PR #468, opened 2026-05-16
[LOOP-413] scanner: liveness heartbeat + watchdog for silent-wedge detection

svv2014:
Closes #413

### `scanner/scanner.sh`
- Added `_write_heartbeat()`: writes the current epoch timestamp to `${LOOP_LOG_DIR}/scanner-heartbeat` at the start of every tick. This gives external observers a clock they can check without reading process internals.
- Added `_check_log_fd()`: at the top of each tick, checks whether `LOG_FILE` is writable. If not (e.g. after log rotation), re-attaches stdout/stderr to the log path. If that fails, exits cleanly so launchd/cron restarts the scanner. Both helpers are called by `run_once()` before any scan logic runs.

- Standalone watchdog script that reads the heartbeat file mtime and compares it to `LOOP_SCANNER_WATCHDOG_STALE` (default: `2 × LOOP_SCANNER_INTERVAL`, i.e. 600s).
- If stale: reads the scanner PID from `/tmp/loop-scanner.lock`, sends `SIGTERM`, waits 2s, then `SIGKILL` if still alive, and removes the lock file so launchd/cron can restart cleanly.
- `--dry-run` flag for manual verification without killing anything.
- On Linux (cron mode), killing is sufficient — cron fires `scanner.sh --once` on the next tick.

- `StartInterval=300` (every 5 minutes). Worst-case detection latency: 5 min (watchdog period) + 10 min (stale threshold) = 15 min, consistent with the issue's target.

### `install.sh`
- `bootstrap_register_launchd`: installs `com.user.loop-scanner-watchdog.plist` alongside the scanner plist.
- `bootstrap_register_cron` (Linux): adds `*/5 * * * * scanner-watchdog.sh` cron entry.

- 8 tests covering: heartbeat creation, timestamp sanity, mtime update on each call, dry-run no-op, presence after `run_once()`, and watchdog dry-run paths (fresh, stale, missing heartbeat).

## Test Plan
- All 8 new tests pass (`bats tests/scanner-heartbeat.bats`).
- `bash -n` and `shellcheck -S warning` pass on all shell files.
- No personal identifiers in committed files.
- Pre-existing test failures in `scanner.bats` (tests 15, 16, 20, 21, 28) and `scanner-jobs-enqueue.bats` (53, 54) are unrelated to this change and present on `main` before this branch.
- Manual simulation: `kill -STOP $SCANNER_PID` then run watchdog with `LOOP_SCANNER_WATCHDOG_STALE=60 --dry-run` to verify stale detection. Remove `--dry-run` and let launchd restart.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -338,6 +338,23 @@ bootstrap_register_launchd() {
         fi
     fi
 
+    local watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
+            -e "s|__HOME__|$HOME|g" \
+            -e "s|__EXTRA_PATH__|$extra_path|g" \
+            "$template_dir/com.user.loop-scanner-watchdog.plist.template" > "$watchdog_plist"
+        if launchctl load "$watchdog_plist" 2>/dev/null; then
+            echo "[launchd] com.user.loop-scanner-watchdog loaded (every 5 min)"
+        else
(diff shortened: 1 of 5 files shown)


svv2014:

PR #471, opened 2026-05-16
[LOOP-413] scanner: heartbeat file + auto-restart watchdog

svv2014:
Closes #413

### `scanner/scanner.sh`
- Defines `HEARTBEAT_FILE="${LOOP_LOG_DIR}/scanner-heartbeat"` at startup.
- In `run_once()`, writes `date +%s` to the heartbeat file at the top of every tick (before the lock sweep, so it fires even on the fastest path). Skip in `--dry-run` mode.

- Sources `lib/env.sh` to pick up `LOOP_LOG_DIR` and `LOOP_SCANNER_INTERVAL`.
- Computes `STALE_THRESHOLD = 2 × POLL_INTERVAL` (default 600 s), overridable via `LOOP_SCANNER_STALE_THRESHOLD`.
- Reads the mtime of the heartbeat file; if absent or older than the threshold logs a warning and restarts:
  - **macOS**: `launchctl kickstart -k gui/<uid>/com.user.loop-scanner`
  - **Linux**: `pkill -f scanner.sh` (cron re-invokes on the next 5-min tick)
- `--dry-run` flag logs actions without executing them.

- Launchd agent for the watchdog. `StartInterval 300` (every 5 min). Uses standard `__LOOP_ROOT__` / `__LOG_DIR__` / `__HOME__` / `__EXTRA_PATH__` substitution tokens. Logs to `loop-scanner-watchdog.log`.

### `install.sh`
- `bootstrap_register_launchd`: registers `com.user.loop-scanner-watchdog` plist alongside existing services.
- `bootstrap_register_cron`: adds `*/5 * * * * restart-scanner-if-stale.sh` cron entry with `# loop-scanner-watchdog` idempotency marker.

- 9 bats tests covering: heartbeat write in scanner source, script existence + syntax, dry-run healthy/absent/stale heartbeat paths, and launchd template presence + `StartInterval` value.

## Test Plan
- `bash -n` and `shellcheck -S warning` pass on all shell files.
- `bats tests/scanner-heartbeat.bats` — all 9 tests pass.
- Manual: kill scanner with `kill -STOP $PID`, wait >10 min, confirm watchdog log shows restart message.
- Verify `${LOOP_LOG_DIR}/scanner-heartbeat` is updated every poll cycle.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,25 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — fires every 5 min; kills and restarts the
+    # scanner if its heartbeat file is older than 2× the poll interval.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
+            -e "s|__HOME__|$HOME|g" \
+            -e "s|__EXTRA_PATH__|$extra_path|g" \
+            "$template_dir/com.user.loop-scanner-watchdog.plist.template" > "$scanner_watchdog_plist"
+        if launchctl load "$scanner_watchdog_plist" 2>/dev/null; then
+            echo "[launchd] com.user.loop-scanner-watchdog loaded (every 5 min)"
+        else
+            echo "[launchd] WARNING: could not load com.user.loop-scanner-watchdog (check Console.app)" >&2
+        fi
+    fi
 }
 
 # Register scanner + reconciler via crontab (Linux). Idempotent: skips if marker present.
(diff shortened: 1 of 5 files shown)


svv2014 (commented):
[loop:review] The heartbeat file + watchdog (AC #1) is solid: cross-platform stat, dry-run support, bats coverage, and idempotent launchd/cron registration all look correct. However, the PR claims `Closes #413` but only satisfies 1 of 3 explicit acceptance criteria in the issue body — AC #2 (emit `scanner_tick` event per tick with `now`+`dedup_count`) and AC #3 (stdout file integrity check per tick) are not implemented. Merging now would silently drop those items when the issue auto-closes.

**Bug — `_heartbeat_age` stat fallback (scripts/restart-scanner-if-stale.sh ~line 50):** the `|| echo 0` fallback sets `mtime=0` when both `stat` variants fail; `age = now - 0 ≈ 1.7B seconds` — always above the stale threshold — causing the watchdog to restart the scanner on every invocation. Change the fallback to `|| echo 999999` and return early (matching the absent-file path) to avoid a restart loop.

**Minor — misleading dry-run log in `_restart_scanner_launchd`:** the `log "restarting scanner via launchctl …"` line fires before the `if $DRY_RUN` guard, so dry-run output claims it is restarting when it is not. Move that log inside the non-dry-run branch.

**Required before merge:** either implement ACs #2 and #3, or change `Closes #413` to a partial-fix note and open follow-up issues for the remaining two items. Fix the `echo 0` stat fallback.

svv2014:

PR #474, opened 2026-05-16
[LOOP-413] scanner: liveness heartbeat + watchdog for silent-stop detection

svv2014:
Closes #413

### `scanner/scanner.sh`
- `run_once()` now calls `touch "${LOOP_LOG_DIR}/scanner-heartbeat"` at the start of every tick (skipped in `--dry-run` mode). This gives the watchdog a reliable mtime to monitor without adding any external dependencies.

- If the file is absent or older than `LOOP_SCANNER_WATCHDOG_STALE_SECS` (default: 900s = 3× the 300s poll interval), the scanner is considered wedged.
- Reads the scanner PID from `/tmp/loop-scanner.lock` and kills it; falls back to `pkill` if the lock is missing.
- launchd `KeepAlive=true` / cron restarts a fresh instance automatically after the kill.
- Supports `--dry-run` for observability without action.

- Runs `scanner-watchdog.sh` every 5 minutes (`StartInterval 300`).

### `install.sh`
- `bootstrap_register_launchd`: registers `com.user.loop-scanner-watchdog` alongside the other agents.
- `bootstrap_register_cron` (Linux): adds a `*/5 * * * *` watchdog cron entry.

- 3 bats tests: heartbeat file is created, mtime updated on successive `run_once` calls, and *not* written in dry-run mode.

## Test Plan
- `bats tests/scanner-heartbeat.bats` — all 3 pass locally.
- Manual wedge simulation: `kill -STOP $SCANNER_PID` → after ≤ 15 min the watchdog detects stale heartbeat and kills the paused process; launchd restarts the scanner.
- `scanner-watchdog.sh --dry-run` reports staleness without killing.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,23 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #475, opened 2026-05-16

svv2014:
Closes #413

### scanner/scanner.sh
- At the top of every `run_once()` tick: writes the current epoch timestamp to the heartbeat file (skipped in `--dry-run` mode)
- Added stdout integrity check: if `$LOG_FILE` is not writable, attempts to reopen via `_scanner_reopen_log` and exits if that fails, letting launchd restart the process

- Standalone watchdog script that runs every 5 minutes
- Reads the heartbeat file mtime; if age > `2 × LOOP_SCANNER_INTERVAL` (default 10 min), the scanner is considered wedged
- Kills the scanner PID (from `/tmp/loop-scanner.lock`) and on macOS uses `launchctl kickstart` to restart immediately; on Linux the next cron tick restarts it
- Supports `--dry-run` flag

- launchd template for the scanner watchdog; fires every 5 minutes via `StartInterval`

### install.sh
- launchd path: registers `com.user.loop-scanner-watchdog` alongside the existing watchdogs
- cron path: adds a `*/5 * * * *` crontab entry for `scanner-watchdog.sh`

- 4 bats tests: heartbeat file is created by `run_once`, timestamp is updated on each call, file holds epoch seconds, `HEARTBEAT_FILE` variable is declared in scanner.sh

## Test Plan
- `bats tests/scanner-heartbeat.bats` passes (4/4)
- `bash -n` and `shellcheck -S warning` clean on all scripts
- Manual test: `kill -STOP $SCANNER_PID` → watchdog detects stale heartbeat within one 5-min interval and kicks the scanner

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,25 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — kills + restarts the scanner if its heartbeat
+    # file goes stale (> 2 × poll interval). Runs every 5 min.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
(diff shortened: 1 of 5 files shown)


svv2014:

PR #476, opened 2026-05-16

svv2014:
Closes #413

### `scanner/scanner.sh`
- Added `_scanner_write_heartbeat()` — writes the current epoch to `${LOOP_LOG_DIR}/scanner-heartbeat` at the start of every real (non-dry-run) tick.
- `run_once()` calls `_scanner_write_heartbeat` as its first action so the file is updated even if a later `scan_project` call hangs.

- Reads the heartbeat file mtime; if age ≥ `LOOP_SCANNER_WATCHDOG_THRESHOLD` (default: `2 × LOOP_SCANNER_INTERVAL` = 600 s), sends SIGTERM to the scanner PID (from `/tmp/loop-scanner.lock`) and on macOS runs `launchctl kickstart -k` so the KeepAlive service restarts immediately.
- On Linux, killing the process is sufficient; cron restarts it on the next `*/5` tick.
- Exits 0 in all cases (safe to run every 5 min without noise when the scanner is healthy).

- `StartInterval 300` (every 5 min) plist template for macOS launchd.

### `install.sh`
- `bootstrap_register_launchd()`: installs `com.user.loop-scanner-watchdog` plist alongside the existing scanner/reconciler/pr-watchdog plists.
- `bootstrap_register_cron()`: adds a `*/5 * * * *` watchdog cron entry for Linux alongside the existing scanner entry.

## Test Plan

- `tests/scanner-heartbeat.bats` (8 new tests):
  - `_scanner_write_heartbeat` creates and updates the heartbeat file with a valid epoch.
  - `run_once()` writes the file in normal mode and skips it in `--dry-run`.
  - Watchdog exits 0 with "no heartbeat file" message when file is absent.
  - Watchdog exits 0 with "heartbeat ok" when file is fresh (10 s old vs 600 s threshold).
  - Watchdog logs "STALE" and proceeds to restart when file mtime > threshold.
- Manual smoke test: `kill -STOP $SCANNER_PID` → watchdog should restart within 10 min.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,15 +354,36 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — restarts a silently wedged scanner.
+    # Fires every 5 min; checks heartbeat mtime; kills+kickstarts if stale.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #480, opened 2026-05-17

svv2014:
Closes #413

### `scanner/scanner.sh`
- Added `HEARTBEAT_FILE` variable (`${LOOP_LOG_DIR}/scanner-heartbeat`).
- `run_once()` now `touch`es the heartbeat file at the start of every tick — even a wedged tick (where `scan_project` hangs) won't update it, so the watchdog can distinguish a live-but-silent scanner from a running one.
- Added a log-file writable check at the top of each tick: if `$LOG_FILE` is not writable (deleted inode after rotation without SIGHUP), `_scanner_reopen_log` is called; if still unwritable, the scanner exits so launchd restarts it with fresh FDs.

- Reads `scanner-heartbeat` mtime; if older than `2 × LOOP_SCANNER_INTERVAL` (default 600 s) logs STALE, kills the scanner PID (from `/tmp/loop-scanner.lock`), removes the stale lock, and on macOS `launchctl kickstart`s the scanner job immediately.
- Under launchd `KeepAlive=true` the kill alone is sufficient — `kickstart -k` just makes it faster.
- `--dry-run` flag prints what would happen without touching anything.

- Runs `scanner-watchdog.sh` every 5 minutes (`StartInterval=300`) via launchd.
- Uses the same `__LOOP_ROOT__` / `__LOG_DIR__` / `__HOME__` / `__EXTRA_PATH__` substitution tokens as the scanner plist so `bootstrap_register_launchd` can install it with one `sed` pass.

### `install.sh`
- `bootstrap_register_launchd()` registers `com.user.loop-scanner-watchdog` after the other plists.
- `bootstrap_register_cron()` adds a `*/5 * * * *` watchdog cron entry (Linux path) alongside the scanner entry.

Six bats tests covering:
- `run_once()` creates heartbeat file on first tick.
- `run_once()` updates heartbeat mtime on subsequent ticks.
- Watchdog exits 0 and is silent when heartbeat is fresh.
- Watchdog exits 0 with informational message when heartbeat file is absent.
- Watchdog logs STALE + DRY-RUN when heartbeat is artificially old.
- Watchdog does not remove the lock file in dry-run mode.

## Test Plan
- `bash -n` and `shellcheck -S warning` — both clean.
- `bats tests/scanner-heartbeat.bats` — 6/6 pass.
- Existing `tests/scanner.bats` — same 5 pre-existing failures, no regressions.
- Manual: `kill -STOP $SCANNER_PID` → within 10 min the watchdog detects stale heartbeat and `kill`s it → launchd restarts.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,26 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog (#413) — kills and restarts the scanner if its
+    # heartbeat file is stale (2×poll_interval). Runs every 5 min on its own
+    # cadence so a wedged KeepAlive scanner is detected within 10 minutes.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
+            -e "s|__HOME__|$HOME|g" \
+            -e "s|__EXTRA_PATH__|$extra_path|g" \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #483, opened 2026-05-17
[LOOP-413] scanner: liveness heartbeat + watchdog for silent-stop detection

svv2014:
Closes #413

## Changes

**`scanner/scanner.sh`**
- Added `HEARTBEAT_FILE` constant (`${LOOP_LOG_DIR}/scanner-heartbeat`)
- Added `_scanner_update_heartbeat()` — writes `now pid=$$` to the heartbeat file at the top of every tick (including `--dry-run`; liveness is independent of dispatch mode)
- Added `_scanner_check_log_fd()` — at the top of every tick, if `LOG_FILE` is absent/unwritable, attempts to reopen stdout/stderr; exits with code 1 on failure so launchd KeepAlive restarts with fresh file descriptors
- Both called at the start of `run_once()`

- Standalone script checking heartbeat mtime; if age > `2 × LOOP_SCANNER_INTERVAL` (default 600s), kills the scanner PID from the lock file and logs
- launchd KeepAlive on the scanner plist auto-restarts within seconds
- Supports `--dry-run`; exits 0 (no-op) when heartbeat file is absent

- Runs `scanner-watchdog.sh` every 5 minutes via `StartInterval`
- Uses same `__LOOP_ROOT__` / `__LOG_DIR__` / `__HOME__` / `__EXTRA_PATH__` substitution pattern

**`install.sh`**
- `bootstrap_register_launchd`: registers `com.user.loop-scanner-watchdog` (idempotent)
- `bootstrap_register_cron`: adds `*/5` crontab entry with `# loop-scanner-watchdog` marker (idempotent)

- `_scanner_update_heartbeat`: creates file, writes pid, updates mtime
- `scanner-watchdog.sh`: absent heartbeat (no-op), fresh heartbeat (ok), stale + dry-run

## Test Plan

- [ ] `bash -n` + `shellcheck -S warning` pass on all scripts
- [ ] `bats tests/scanner-heartbeat.bats` — all 6 green
- [ ] No regressions in existing scanner test suite
- [ ] Manual: inspect `${LOOP_LOG_DIR}/scanner-heartbeat` updates each tick
- [ ] Manual: `scanner-watchdog.sh --dry-run` reports "ok:" when heartbeat fresh
- [ ] Manual: `kill -STOP $SCANNER_PID` → watchdog kills+restarts within 10 min

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,25 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — checks heartbeat file every 5 min and kills
+    # a wedged scanner so launchd KeepAlive restarts it automatically.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
+            -e "s|__HOME__|$HOME|g" \
+            -e "s|__EXTRA_PATH__|$extra_path|g" \
+            "$template_dir/com.user.loop-scanner-watchdog.plist.template" > "$scanner_watchdog_plist"
+        if launchctl load "$scanner_watchdog_plist" 2>/dev/null; then
+            echo "[launchd] com.user.loop-scanner-watchdog loaded (every 5 min)"
+        else
+            echo "[launchd] WARNING: could not load com.user.loop-scanner-watchdog (check Console.app)" >&2
+        fi
+    fi
 }
 
 # Register scanner + reconciler via crontab (Linux). Idempotent: skips if marker present.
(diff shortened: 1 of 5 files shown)


svv2014 (commented):
[loop:review] REQUEST_CHANGES — Two substantive gaps prevent this from fully meeting the issue spec:

**1. Missing acceptance criterion #2 — `scanner_tick` event not emitted.**
The issue explicitly requires: "every tick emit a `scanner_tick` event with `now` + `dedup_count` to loop-monitor. Monitor surfaces it on the dashboard so silent-stop is visible." The PR implements heartbeat (AC#1) and stdout-fd recovery (AC#3) but omits the tick-counter event entirely. This is load-bearing: without it, a wedged scanner is still invisible to the dashboard until the watchdog fires (up to 10 min gap). Add a `scanner_tick` emit call in `run_once()` after `_scanner_update_heartbeat`, guarded by the existing `BOBA_EVENT_CLIENT` check pattern already used elsewhere in the scanner.

**2. Logic bug in `_scanner_check_log_fd` — `&&` should be `||`.**
Current: `if [ ! -w "$LOG_FILE" ] && [ ! -e "$LOG_FILE" ]`
Since `! -e` ⊆ `! -w` (a missing file is never writable), this condition is equivalent to just `[ ! -e "$LOG_FILE" ]`. It silently skips the case where the log file exists but loses write permission — exactly the failure mode described in the issue ("StandardOutPath was 0 bytes since May 14"). Fix: `if [ ! -w "$LOG_FILE" ] || [ ! -e "$LOG_FILE" ]` (or simply `if [ ! -w "$LOG_FILE" ]`).

Everything else looks solid: heartbeat write, watchdog threshold math, launchd plist template, idempotent `install.sh` registration, and the 6-case bats suite are all correct and well-structured.

svv2014:

PR #485, opened 2026-05-17
[LOOP-413] scanner: add liveness heartbeat + watchdog for silent-stop detection

svv2014:
Closes #413

### `scanner/scanner.sh`
- `run_once()` now calls `touch "${LOOP_LOG_DIR}/scanner-heartbeat"` at the start of every tick (skipped in `--dry-run` mode). The watchdog reads this file's mtime to determine if the scanner is alive.
- Added a log-FD integrity check at the top of each tick: if `LOG_FILE` exists but is no longer writable (e.g. log rotation without SIGHUP), the FDs are reopened. If reopening fails, the process exits so launchd/cron restarts a fresh one. This complements the existing SIGHUP handler which handles the signal-based rotation path.

- If the file is older than `LOOP_WATCHDOG_STALE_SECONDS` (default 600 s = 2× the 5-min poll interval), kills the stale scanner PID (read from the lock file) and:
  - macOS: calls `launchctl kickstart -k gui/<uid>/com.user.loop-scanner`
  - Linux: kills the PID; cron restarts it on the next `*/5` interval.
- Supports `--dry-run`.

- Runs `scanner-watchdog.sh` every 5 minutes (`StartInterval=300`).
- Same placeholder scheme as other launchd templates (`__LOOP_ROOT__`, `__LOG_DIR__`, `__HOME__`, `__EXTRA_PATH__`).

### `install.sh`
- `bootstrap_register_launchd()`: registers `com.user.loop-scanner-watchdog` alongside the existing scanner/reconciler/pr-watchdog plists.
- `bootstrap_register_cron()`: adds a `*/5` cron entry for `scanner-watchdog.sh` alongside the scanner entry (Linux path).

Three bats tests:
1. `run_once` creates the heartbeat file on first tick.
2. `run_once` updates the heartbeat mtime on each subsequent call.
3. Heartbeat is **not** written in `--dry-run` mode.

## Test Plan
- All three new bats tests pass (`bats tests/scanner-heartbeat.bats`).
- Existing scanner tests pass with no new failures.
- `bash -n` and `shellcheck -S warning` clean on all shell files.
- Manual simulation: `kill -STOP $SCANNER_PID` → watchdog kills and restarts within 10 min (2× poll interval watchdog window).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -338,6 +338,23 @@ bootstrap_register_launchd() {
         fi
     fi
 
+    local watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
+            -e "s|__HOME__|$HOME|g" \
+            -e "s|__EXTRA_PATH__|$extra_path|g" \
+            "$template_dir/com.user.loop-scanner-watchdog.plist.template" > "$watchdog_plist"
+        if launchctl load "$watchdog_plist" 2>/dev/null; then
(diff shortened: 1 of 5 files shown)


svv2014:

PR #487, opened 2026-05-17
[LOOP-413] scanner: heartbeat file + watchdog auto-restart

svv2014:
Closes #413

## Changes

**Problem:** The scanner process could become silently wedged — alive PID, sleeping between ticks, but emitting zero events for hours. No alert, no telemetry. The only signal was comparing `loop-scanner.log` mtime to now.

**Solution:**

1. **Heartbeat file** (`scanner.sh`): `run_once()` now `touch`es `${LOOP_LOG_DIR}/scanner-heartbeat` at the start of every tick. This gives the watchdog a reliable, per-tick liveness signal.

2. **Watchdog script** (`scanner/scanner-watchdog.sh`): Runs independently every 5 min. Reads the heartbeat file's mtime; if age exceeds `LOOP_SCANNER_WATCHDOG_THRESHOLD` (default: `2 × POLL_INTERVAL + 60 s = 660 s`, configurable), it considers the scanner wedged, sends `SIGTERM` to the scanner PID from the lock file, and on Linux directly spawns a replacement (macOS relies on launchd `KeepAlive`).

3. **launchd template** (`templates/launchd/com.user.loop-scanner-watchdog.plist.template`): `StartInterval=300` (5 min), installed alongside the existing scanner plist.

4. **install.sh**: Registers `com.user.loop-scanner-watchdog` on macOS via `bootstrap_register_launchd` and adds a `*/5 * * * *` crontab entry on Linux via `bootstrap_register_cron`.

5. **Tests** (`tests/scanner-heartbeat.bats`): 4 bats tests — source guard that heartbeat write is present, healthy exit, missing-file exit, and stale-heartbeat ALERT path.

## Test Plan

- [ ] All 4 bats tests pass: `bats tests/scanner-heartbeat.bats`
- [ ] `bash -n` and `shellcheck -S warning` pass on all scripts
- [ ] Manual: `kill -STOP $SCANNER_PID` → watchdog detects stale heartbeat within 10 min and kills it → launchd restarts scanner
- [ ] After install: verify `~/Library/LaunchAgents/com.user.loop-scanner-watchdog.plist` exists and `launchctl list | grep loop-scanner-watchdog` shows it loaded

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,25 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — kills a wedged scanner (alive PID, no events)
+    # so launchd auto-restarts it. Runs every 5 min; harmless when scanner is healthy.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
+            -e "s|__HOME__|$HOME|g" \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #489, opened 2026-05-17

svv2014:
Closes #413

### scanner/scanner.sh
- Added `touch ${LOOP_LOG_DIR}/scanner-heartbeat` at the start of every `run_once()` tick (skipped in dry-run mode).
- Added a log-file writability check at the start of each tick: if the log FD is broken, re-exec FDs 1+2 against the log path or exit so launchd can restart cleanly.

### scripts/check-scanner-liveness.sh (new)
- Watchdog script run every 5 minutes (via launchd or cron).
- Reads `${LOOP_LOG_DIR}/scanner-heartbeat` mtime. If age > `LOOP_SCANNER_STALE_THRESHOLD` seconds (default 600 = 2× default poll interval), the scanner is considered wedged.
- Reads PID from `/tmp/loop-scanner.lock`, sends SIGTERM (then SIGKILL after 2s), then calls `launchctl kickstart` on macOS so the scanner restarts immediately rather than waiting for the next ThrottleInterval.
- Always exits 0 so launchd does not penalise the watchdog itself.

- `StartInterval=300` (every 5 min). `RunAtLoad=false` to avoid a spurious kill before the first scanner tick writes the heartbeat.

### install.sh
- `bootstrap_register_launchd()`: registers `com.user.loop-scanner-watchdog` plist (idempotent).
- `bootstrap_register_cron()`: adds `*/5 * * * *` cron entry for `check-scanner-liveness.sh` on Linux.

6 bats tests covering:
1. `run_once` creates heartbeat on first tick.
2. `run_once` updates heartbeat mtime on subsequent ticks.
3. `run_once` does NOT write heartbeat in dry-run mode.
4. Watchdog exits 0 and reports "live" when heartbeat is fresh.
5. Watchdog exits 0 even when heartbeat is stale and no scanner PID is found.
6. Watchdog kills the wedged PID when heartbeat is stale and a live PID exists.

## Test Plan
- All 6 new bats tests pass (`bats tests/scanner-heartbeat.bats`).
- Manual simulation: `kill -STOP $SCANNER_PID` → heartbeat goes stale → watchdog (next 5-min tick) kills and kickstarts the scanner.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,25 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — checks heartbeat file every 5 min and
+    # kills/restarts the scanner if it has been silent for > 2× poll interval.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
+            -e "s|__HOME__|$HOME|g" \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #490, opened 2026-05-17

svv2014:
Closes #413

### `scanner/scanner.sh`
- `run_once()` now calls `touch "${LOOP_LOG_DIR}/scanner-heartbeat"` at the start of every tick (skipped in `--dry-run` mode). Any external process can detect silent-stop by comparing the heartbeat file mtime against now.
- Added stdout integrity check: if `LOG_FILE` exists but is no longer writable (e.g., after logrotate without SIGHUP), the scanner attempts to reopen it; if that also fails it exits immediately so launchd restarts it with a clean fd table.

- Single-shot script designed to run every 5 minutes via launchd or cron.
- Reads `${LOOP_LOG_DIR}/scanner-heartbeat` mtime. If older than `LOOP_SCANNER_STALE_THRESHOLD` seconds (default: `2 × LOOP_SCANNER_INTERVAL` = 600 s), the scanner is considered wedged.
- Kills the scanner PID (from `/tmp/loop-scanner.lock`) and calls `launchctl kickstart` on macOS so the restart is immediate rather than waiting for `ThrottleInterval`.
- Safe `--dry-run` flag for testing without side effects.

- Runs `scanner-watchdog.sh` every 5 minutes (`StartInterval=300`).

### `install.sh`
- `bootstrap_register_launchd`: registers the scanner-watchdog plist alongside the existing scanner/reconciler plists.
- `bootstrap_register_cron`: adds a `*/5 * * * *` cron entry for `scanner-watchdog.sh` on Linux.

- 6 tests covering: heartbeat file created/updated on tick, skipped in dry-run, watchdog detects fresh/stale/absent heartbeat.

## Test Plan
- All 6 `scanner-heartbeat.bats` tests pass locally.
- No personal identifiers detected.
- Manual: simulate wedge via `kill -STOP $SCANNER_PID` → watchdog should log "WARN: scanner stale" and trigger restart within 10 minutes (2 × poll interval).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,26 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — fires every 5 min to check the heartbeat file
+    # written by scanner.sh. If the scanner is silently wedged, kills its PID so
+    # launchd (KeepAlive=true) restarts it automatically.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #492, opened 2026-05-17

svv2014:
Closes #413

## Changes

**`scanner/scanner.sh`** — touch `${LOOP_LOG_DIR}/scanner-heartbeat` at the start of every `run_once()` tick. This gives an always-current mtime signal that reflects whether the scan loop is alive.

**`scripts/scanner-watchdog.sh`** (new) — runs every 5 minutes. Reads the heartbeat file mtime; if age > `2 × LOOP_SCANNER_INTERVAL + 60s` (default 660s ≈ 11 min), the scanner is considered wedged: its PID (from the lock file) is killed so launchd (KeepAlive=true) auto-restarts it. On macOS, `launchctl kickstart` is attempted for an immediate restart rather than waiting out the 60s ThrottleInterval.

**`templates/launchd/com.user.loop-scanner-watchdog.plist.template`** (new) — launchd plist for the watchdog; `StartInterval=300` (5 min), no `KeepAlive`.

**`install.sh`** — registers the watchdog plist in `bootstrap_register_launchd()` and adds a `*/5` cron entry in `bootstrap_register_cron()` for Linux.

**`tests/scanner-heartbeat.bats`** (new) — 5 bats tests covering:
- `run_once` creates heartbeat on first tick
- `run_once` updates heartbeat mtime on subsequent ticks
- watchdog reports healthy when heartbeat is fresh
- watchdog reports stale (dry-run) when heartbeat is old
- watchdog exits cleanly when heartbeat file is absent

## Test Plan

- All 846 existing bats tests pass
- New `tests/scanner-heartbeat.bats` (5 tests) all pass
- Manual: `kill -STOP $SCANNER_PID` → watchdog kills it within 11 min → launchd restarts
- `scanner-watchdog.sh --dry-run` shows healthy/stale without side effects

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,24 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — kills a wedged scanner so launchd auto-restarts it.
+    local watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #494, opened 2026-05-17
[LOOP-413] scanner: write heartbeat file each tick + watchdog restart script

svv2014:
Closes #413

## Changes

- **scanner/scanner.sh**: Added `_scanner_write_heartbeat()` which writes the current UTC timestamp to `${LOOP_LOG_DIR}/scanner-heartbeat` at the start of every tick (skipped in dry-run). Added `_scanner_check_log_fd()` which exits with status 1 when the log file is no longer writable, allowing launchd to restart the process with a fresh file descriptor.

- **scripts/restart-scanner-if-stale.sh**: New watchdog script that reads the heartbeat file mtime. If older than `LOOP_SCANNER_STALE_THRESHOLD` seconds (default 900 / 15 min), it sends SIGTERM to the scanner PID (from the lock file), then SIGKILL if needed, and calls `launchctl kickstart` as a belt-and-suspenders fallback. Falls back to checking the scanner log when the heartbeat file doesn't exist yet. Supports `--dry-run`.

- **templates/launchd/com.user.loop-scanner-watchdog.plist.template**: launchd plist template that fires `restart-scanner-if-stale.sh` every 5 minutes (StartInterval=300). Follows the same `@LOOP_ROOT@`/`@LOG_DIR@` substitution pattern as other plist templates.

- **tests/scanner-heartbeat.bats**: bats test suite covering: heartbeat file creation on first tick, ISO timestamp format validation, mtime update on repeat, dry-run no-op, log fd check pass/fail, watchdog healthy path, watchdog stale path (backdated mtime), dry-run no-kill, and absent-file graceful skip.

## Test Plan

- Run `bats tests/scanner-heartbeat.bats` — all tests should pass
- Manual: start scanner, verify `${LOOP_LOG_DIR}/scanner-heartbeat` is updated each tick
- Manual: install watchdog plist, then `kill -STOP $SCANNER_PID` — watchdog should detect stale heartbeat within 10 min and restart scanner
- Verify `--dry-run` mode prints action without killing anything

### DIFF
--- a/scanner/scanner.sh
+++ b/scanner/scanner.sh
@@ -29,6 +29,7 @@ source "$LOOP_ROOT/lib/jobs.sh"
 
 LOCK_FILE="/tmp/loop-scanner.lock"
 LOG_FILE="${LOOP_LOG_DIR}/loop-scanner.log"
+HEARTBEAT_FILE="${LOOP_LOG_DIR}/scanner-heartbeat"
 POLL_INTERVAL="${LOOP_SCANNER_INTERVAL:-300}"
 BOBA_EVENT_CLIENT="${LOOP_EVENT_CLIENT:-}"
 HANDLER_TIMEOUT="${LOOP_HANDLER_TIMEOUT:-7200}"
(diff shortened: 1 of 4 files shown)


svv2014 (commented):
[loop:review] Heartbeat write + log-fd check in `run_once`, watchdog script, launchd template, and bats suite all look correct and well-scoped to the "Alternative simpler implementation" path in the issue. Two minor nits to be aware of: (1) `date '+%Y-%m-%dT%H:%M:%SZ'` in `_scanner_write_heartbeat` lacks the `-u` flag — writes local time with a misleading Z suffix (harmless since the watchdog only reads `stat` mtime, not file content); (2) `tee -a "$WATCHDOG_LOG"` combined with the plist `StandardOutPath` pointing to the same file will double every log line in the watchdog log. Acceptance criterion #2 (tick-counter event) is intentionally deferred per the issue own alternative path; a follow-up issue should track it. LGTM — advancing to QA.

svv2014:

PR #498, opened 2026-05-17

svv2014:
Closes #413

### Problem addressed
The scanner process could become silently wedged — alive PID, sleep loop intact, but emitting no events for hours. No alert, no telemetry. The only way to detect it was to compare `loop-scanner.log` mtime against now.

### Implementation

**Heartbeat write (scanner.sh)**
At the top of every `run_once()` tick, the scanner writes `${LOOP_LOG_DIR}/scanner-heartbeat` with the current unix timestamp and PID. This file is not written in `--dry-run` mode.

**Watchdog script (scripts/scanner-watchdog.sh)**
New script intended to run every 5 min. Reads the heartbeat file mtime; if older than `LOOP_WATCHDOG_STALE_SECONDS` (default: `2 × LOOP_SCANNER_INTERVAL = 600 s`), it kills the scanner PID from `/tmp/loop-scanner.lock` (SIGTERM, then SIGKILL after 5 s). launchd `KeepAlive=true` restarts the scanner automatically within seconds.

Supports `--dry-run` for safe inspection.

**launchd plist template (macOS)**
`templates/launchd/com.user.loop-scanner-watchdog.plist.template` — `StartInterval=300`, registered by `install.sh --bootstrap`.

**cron entry (Linux)**
`bootstrap_register_cron()` in `install.sh` now also adds a `*/5` cron entry for `scanner-watchdog.sh`.

**Tests**
`tests/scanner-heartbeat.bats` — 6 tests:
- Heartbeat file created on first tick
- File contains valid unix timestamp and PID
- Mtime advances on second tick
- File NOT written in dry-run mode
- Watchdog exits 0 with "ok" message when heartbeat is fresh
- Watchdog reports STALE when heartbeat is old (dry-run mode)

## Test Plan

- [ ] `bash -n` syntax check passes on all scripts
- [ ] `shellcheck -S warning` passes
- [ ] No personal identifiers in committed files
- [ ] All 6 new bats tests pass (`bats tests/scanner-heartbeat.bats`)
- [ ] Manual: after `install.sh --bootstrap`, confirm `com.user.loop-scanner-watchdog` appears in `launchctl list`
- [ ] Manual: simulate wedge with `kill -STOP $SCANNER_PID`; within 10 min watchdog should kill and launchd should restart

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -338,6 +338,26 @@ bootstrap_register_launchd() {
         fi
     fi
 
+    # Scanner liveness watchdog — kills + lets launchd restart a wedged scanner.
+    # Fires every 5 min; compares scanner-heartbeat mtime against the stale
+    # threshold (2 × poll interval). Companion to the KeepAlive scanner plist.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
+            -e "s|__HOME__|$HOME|g" \
+            -e "s|__EXTRA_PATH__|$extra_path|g" \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #499, opened 2026-05-17

svv2014:
Closes #413

## Changes

- **`scanner/scanner.sh`**: `_write_heartbeat()` writes the current timestamp to `${LOOP_LOG_DIR}/scanner-heartbeat` at the start of every tick via `run_once()`. No-op in dry-run mode.

- **`scanner/scanner-watchdog.sh`** (new): Runs every 5 min via launchd. Reads `scanner-heartbeat` mtime; if silent for `> 2 × LOOP_SCANNER_INTERVAL` (default 10 min), kills the wedged scanner PID from the lock file so launchd (KeepAlive=true) auto-restarts it. On macOS, uses `launchctl kickstart` for immediate restart rather than waiting for ThrottleInterval.

- **`templates/launchd/com.user.loop-scanner-watchdog.plist.template`** (new): Launchd plist template for the watchdog. StartInterval=300 (5 min). Follows the same `__LOOP_ROOT__` / `__LOG_DIR__` / `__HOME__` substitution pattern as existing plist templates.

- **`tests/scanner-heartbeat.bats`** (new): 7 tests — 4 for `_write_heartbeat` (creates file, updates mtime, correct timestamp format, no-op in dry-run) and 3 for `scanner-watchdog.sh` (fresh heartbeat → alive, stale heartbeat → wedge + DRY-RUN, missing heartbeat → wedge).

## Test Plan

- `bash -n` + `shellcheck -S warning` — both pass (all scripts)
- `bats tests/scanner-heartbeat.bats` — 7/7 pass
- Existing `bats tests/scanner.bats` failures are pre-existing (label renames) and unrelated to this change
- Manual: `kill -STOP $SCANNER_PID` for 10+ min → watchdog logs "wedged" and kills/restarts; `kill -CONT` before watchdog fires → watchdog logs "alive"

### DIFF
--- a/scanner/scanner-watchdog.sh
+++ b/scanner/scanner-watchdog.sh
@@ -1,0 +1,99 @@
+#!/usr/bin/env bash
+# scanner-watchdog.sh — restart the scanner if it stops emitting heartbeats.
+#
+# Reads ${LOOP_LOG_DIR}/scanner-heartbeat (written every tick by scanner.sh).
+# If the file is missing or its mtime exceeds 2 × POLL_INTERVAL (default 10 min),
+# the scanner is considered wedged: kill the PID from /tmp/loop-scanner.lock so
+# launchd (KeepAlive=true) or cron restarts it.
+#
+# Designed to run every 5 min via launchd StartInterval or cron.
+# Linux fallback: if launchctl is absent, kill the PID and let cron or the
+# process supervisor restart it independently.
+#
+# Flags:
+#   --dry-run   print verdict without killing
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
+LOOP_ROOT="$(cd "$SCRIPT_DIR/.." && pwd)"
+# shellcheck source=../lib/env.sh
+source "$LOOP_ROOT/lib/env.sh"
+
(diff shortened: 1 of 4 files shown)


svv2014 (commented):
[loop:review] The heartbeat file + watchdog logic is solid — lock file path correctly matches `scanner.sh` (/tmp/loop-scanner.lock), cross-platform `stat` fallback works, launchctl kickstart chain is well-structured, and the 7 bats tests cover the key paths. One functional gap needs fixing before merge: **`install.sh` does not register the new `com.user.loop-scanner-watchdog.plist`**. Every other launchd agent (scanner, reconciler, pr-watchdog) is wired in via `install_launchd()` during `./install.sh --bootstrap`; the omission means the watchdog is never active on any existing or new install. The fix follows the exact same pattern as `com.user.loop-pr-watchdog.plist` (lines 324-333 of install.sh). Two minor notes: (1) `_file_age_seconds` stat failure fallback is `echo 0`, which produces age ≈ current epoch (billions of seconds) — harmless but confusing; `echo 999999` would be more consistent with the missing-file path. (2) Acceptance criteria 2 (tick-counter event) and 3 (stdout integrity check) are unimplemented, but the "Alternative simpler implementation" clause in the issue body scopes those out — they can be separate issues. Route-back: adding `loop:action:dev` for the install.sh wiring.

svv2014:

PR #502, opened 2026-05-17

svv2014:
Closes #413

### `scanner/scanner.sh`
- Write `touch ${LOOP_LOG_DIR}/scanner-heartbeat` at the start of every `run_once()` tick. This gives the watchdog a reliable mtime-based liveness signal — if the file goes stale, the scanner is either wedged in a tick or stuck in the sleep loop with a broken state.

### `scanner/restart-scanner-if-stale.sh` (new)
- Sources `lib/env.sh` for `LOOP_LOG_DIR` and env config.
- Reads mtime of `scanner-heartbeat`. If older than `LOOP_SCANNER_WATCHDOG_STALE_SECONDS` (default 900 s = 3× the default 300 s poll interval), kills the PID recorded in `/tmp/loop-scanner.lock`.
- On macOS, issues `launchctl kickstart -k gui/<uid>/com.user.loop-scanner` so KeepAlive fires immediately rather than waiting for the next restart cycle.
- On Linux, killing the PID is sufficient — cron relaunches within the next 5-minute slot.
- Supports `--dry-run` for manual inspection.

- `StartInterval 300` (every 5 min), `RunAtLoad false` (no false-positive on fresh install).
- Uses the same `__LOOP_ROOT__` / `__LOG_DIR__` / `__HOME__` / `__EXTRA_PATH__` substitution pattern as all other templates. Logs to `loop-scanner-watchdog.log`.

### `install.sh`
- `bootstrap_register_launchd()`: installs and loads `com.user.loop-scanner-watchdog.plist` after the digest agent.
- `bootstrap_register_cron()`: adds `*/5 * * * *` watchdog cron entry alongside the existing scanner and reconciler entries.

## Test Plan
- Simulate wedge on macOS: `kill -STOP $SCANNER_PID` → within 15 min the watchdog should log "STALE" and restart the scanner.
- Dry-run: `scanner/restart-scanner-if-stale.sh --dry-run` with a stale heartbeat should print "DRY-RUN: would restart" without touching the process.
- Fresh install: `./install.sh --bootstrap` should register four launchd agents including `com.user.loop-scanner-watchdog`.
- Heartbeat presence: after the scanner runs one tick, `${LOOP_LOG_DIR}/scanner-heartbeat` should exist and have a recent mtime.
- No false positives: during normal operation (heartbeat < 900 s old) the watchdog exits with "heartbeat ok".

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,25 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — kills a wedged scanner process and lets launchd
+    # restart it. Fires every 5 min; acts only when heartbeat is stale (>900 s).
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
+            -e "s|__HOME__|$HOME|g" \
+            -e "s|__EXTRA_PATH__|$extra_path|g" \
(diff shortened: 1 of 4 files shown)


svv2014:

PR #504, opened 2026-05-17

svv2014:
Closes #413

### `scanner/scanner.sh`
- At the top of `run_once()`, writes `touch "$LOOP_LOG_DIR/scanner-heartbeat"` on every tick (skipped in `--dry-run` mode). This gives an external watchdog a reliable mtime signal.

- Runs every 5 min via launchd/cron.
- Reads `$LOOP_LOG_DIR/scanner-heartbeat` mtime.
- Stale threshold: `2 × LOOP_SCANNER_INTERVAL` (default 600 s — two missed ticks).
- If stale: reads the scanner PID from `/tmp/loop-scanner.lock`, sends SIGTERM (then SIGKILL after 2 s if still alive), and calls `launchctl kickstart` so launchd restarts the process immediately rather than waiting out `ThrottleInterval`.
- `--dry-run` flag for safe testing without actual kills.

- `StartInterval 300` (every 5 min), `RunAtLoad false`.
- Same placeholder substitution pattern as existing plists (`__LOOP_ROOT__`, `__LOG_DIR__`, `__HOME__`, `__EXTRA_PATH__`).

### `install.sh`
- `bootstrap_register_launchd`: installs `com.user.loop-scanner-watchdog` alongside the scanner plist (idempotent, skips if already registered).
- `bootstrap_register_cron` (Linux): adds `*/5 * * * *` cron entry for the watchdog (marker-guarded, idempotent).

- Verifies `run_once()` creates the heartbeat file.
- Verifies the mtime is updated on subsequent calls.
- Verifies heartbeat is NOT written in `--dry-run` mode.

## Test Plan
- `bash -n` + `shellcheck -S warning` pass on all scripts (validated locally).
- No personal identifiers.
- Manual: `kill -STOP $SCANNER_PID` → watchdog should detect stale heartbeat within 10 min and restart.
- Bats: `bats tests/scanner-heartbeat.bats`

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -338,6 +338,26 @@ bootstrap_register_launchd() {
         fi
     fi
 
+    # Scanner liveness watchdog — detects a wedged scanner (heartbeat stale)
+    # and kills it so launchd (KeepAlive) auto-restarts within ThrottleInterval.
+    # Runs every 5 min; safe to leave running alongside the scanner at all times.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
+            -e "s|__HOME__|$HOME|g" \
+            -e "s|__EXTRA_PATH__|$extra_path|g" \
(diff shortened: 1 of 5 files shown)


svv2014 (commented):
[loop:review] Heartbeat file + watchdog implementation is solid: portable `stat` mtime fallback, idempotent launchd/cron install, `--dry-run` flag, and three bats tests covering write, mtime-update, and dry-run suppression. CI passes (lint + check-closes-ref green). Items 2 (tick-counter event) and 3 (stdout file integrity check) from the ## Acceptance section are unimplemented, but the issue body explicitly offers the "Alternative simpler implementation" path — this PR takes that path cleanly with a dedicated heartbeat file rather than relying on log mtime, which is a better design. LGTM — advancing to QA.

svv2014:

PR #505, opened 2026-05-17
[LOOP-413] scanner: heartbeat file + external watchdog for silent-stop detection

svv2014:
Closes #413

### scanner/scanner.sh
- Added `_scanner_write_heartbeat()` — writes current unix timestamp to `${LOOP_LOG_DIR}/scanner-heartbeat` at the start of every tick (skipped in `--dry-run` mode). An external watchdog reads this file; stale mtime means the scanner is wedged.
- Added `_scanner_check_log_writable()` — at the start of each tick, verifies the log file is still writable. If the inode has been rotated and is no longer writable, reopens FDs 1+2 against the path; exits so launchd restarts cleanly if reopen fails.
- Extended the `=== scan tick start ===` log line to include `ts=<epoch>` and `dedup_count=<n>` so silent-stop is visible in logs without a separate dashboard.

- Standalone watchdog script that fires every 5 minutes (via launchd or cron).
- Reads `scanner-heartbeat` mtime; if age > `2 × LOOP_SCANNER_INTERVAL` (default 600s) the scanner is considered wedged.
- Kills the PID recorded in the lock file; on macOS also calls `launchctl kickstart` on the scanner service so it restarts immediately rather than waiting for `ThrottleInterval`.

- launchd plist template for the watchdog, firing every 300s (`StartInterval`).
- Follows the same placeholder convention (`__LOOP_ROOT__`, `__LOG_DIR__`, `__HOME__`, `__EXTRA_PATH__`) as the existing scanner plist template.

- 8 bats tests covering: heartbeat file creation, timestamp content, mtime update, dry-run skip, heartbeat written by `run_once()`, and all three watchdog scenarios (fresh / missing / stale heartbeat).

## Test Plan
- `bats tests/scanner-heartbeat.bats` — all 8 tests pass.
- `bash -n` + `shellcheck -S warning` on all `lib/*.sh scripts/*.sh scanner/*.sh install.sh` — clean.
- Manual: simulate wedge with `kill -STOP $SCANNER_PID`; watchdog should trigger restart within 10 min (2 × 300s threshold + up to 300s watchdog cadence).

### DIFF
--- a/scanner/scanner-watchdog.sh
+++ b/scanner/scanner-watchdog.sh
@@ -1,0 +1,84 @@
+#!/usr/bin/env bash
+# scanner-watchdog.sh — restart the scanner if it stops emitting heartbeats.
+#
+# Run every 5 minutes by launchd (macOS) or cron (Linux).
+# If the scanner heartbeat file is older than 2×POLL_INTERVAL the scanner is
+# considered wedged: we kill the PID in the lock file and let launchd/cron
+# restart it via KeepAlive=true.
+#
+# Usage:
+#   scanner-watchdog.sh           # normal run
+#   scanner-watchdog.sh --dry-run # print verdict without killing
+
+set -euo pipefail
+
+SCRIPT_DIR="$(cd "$(dirname "$0")" && pwd)"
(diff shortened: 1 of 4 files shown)


svv2014:

PR #507, opened 2026-05-17
[LOOP-413] scanner: add liveness heartbeat + watchdog auto-restart

svv2014:
Closes #413

## Changes
- `scanner/scanner.sh`: write a UTC ISO-8601 timestamp to `${LOOP_LOG_DIR}/scanner-heartbeat` on every tick. This is the observable liveness signal the watchdog consumes.
- `scripts/restart-scanner-if-stale.sh`: watchdog script — reads heartbeat mtime; if the file is absent or older than `STALE_THRESHOLD` seconds (default 900 s = 3× the 300 s poll interval) it restarts the scanner. macOS path: `launchctl kickstart -k gui/<uid>/<label>` with fallback to stop/start. Linux path: `systemctl restart loop-scanner` with fallback to kill-by-lock-PID + nohup respawn.
- `templates/launchd/com.user.loop-scanner-watchdog.plist.template`: launchd agent that runs the watchdog every 300 s (5 min) via `StartInterval`. Follows the same `__LOOP_ROOT__` / `__LOG_DIR__` / `__HOME__` substitution conventions as the existing scanner and reconciler plists.

## Test Plan
- Manual smoke: run scanner, `kill -STOP $SCANNER_PID`, wait for heartbeat to go stale, run `restart-scanner-if-stale.sh` — confirm it logs the restart
- Verify heartbeat file is written on each tick in continuous and `--once` modes (it is skipped in `--dry-run` mode intentionally)
- `bash -n` and `shellcheck -S warning` pass on all scripts (verified locally before this PR)

### DIFF
--- a/scanner/scanner.sh
+++ b/scanner/scanner.sh
@@ -776,6 +776,8 @@ scan_project() {
 
 run_once() {
     log "=== scan tick start ==="
+    # Write heartbeat so the watchdog can detect a silent-stop scanner.
+    $DRY_RUN || printf '%s\n' "$(date -u +%Y-%m-%dT%H:%M:%SZ)" > "${LOOP_LOG_DIR}/scanner-heartbeat"
     $DRY_RUN || _sweep_stale_locks
     if [[ "${LOOP_JOBS_ENQUEUE:-1}" == "1" ]] && ! $DRY_RUN; then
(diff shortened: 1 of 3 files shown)


svv2014:

PR #508, opened 2026-05-17

svv2014:
Closes #413

## Changes

- **`scanner/scanner.sh`**: write epoch timestamp to `${LOOP_LOG_DIR}/scanner-heartbeat` on every tick (skipped in `--dry-run`). Added a log-file integrity guard that exits the process if the log FD becomes non-writable, allowing launchd to restart cleanly.
- **`scanner/scanner-watchdog.sh`** (new): reads heartbeat mtime; if the file is absent or older than `LOOP_SCANNER_STALE_THRESHOLD` seconds (default 900 s = 3× the 300 s poll interval), kills the wedged PID and restarts via `launchctl kickstart` (macOS) or cron respawn (Linux).
- **`templates/launchd/com.user.loop-scanner-watchdog.plist.template`** (new): launchd agent that runs the watchdog every 5 minutes via `StartInterval 300`. Follows the same `__LOOP_ROOT__`/`__LOG_DIR__`/`__HOME__` substitution convention as existing templates.
- **`install.sh`**: registers the watchdog plist on macOS bootstrap; adds a `*/5 * * * *` cron entry on Linux.
- **`tests/scanner-heartbeat.bats`** (new): bats tests verify the heartbeat file is created, contains a valid unix timestamp, is updated on each tick, and is NOT written in dry-run mode.

## Test Plan

- `bash -n` and `shellcheck -S warning` pass on all scripts (verified locally)
- Run `bats tests/scanner-heartbeat.bats` — all 4 tests pass
- Manual smoke: run scanner, `kill -STOP $SCANNER_PID` to wedge it, wait for heartbeat to go stale, run `scanner/scanner-watchdog.sh` — confirm it logs "STALE" and restarts
- Tune `LOOP_SCANNER_STALE_THRESHOLD` env var to lower value for faster testing

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,25 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — fires every 5 min, restarts scanner if
+    # heartbeat file is stale (default threshold: 15 min / 900 s).
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #510, opened 2026-05-17

svv2014:
Closes #413

### `scanner/scanner.sh`
- `run_once()` now calls `touch "${LOOP_LOG_DIR}/scanner-heartbeat"` at the start of every tick (skipped in `--dry-run`). The file mtime serves as a liveness indicator — if it stops advancing, the scanner is wedged.

- Separate script invoked every 5 minutes by launchd (macOS) or cron (Linux).
- Reads the heartbeat file mtime; if `age >= LOOP_SCANNER_STALE_SECONDS` (default 900 s = 3× the 300 s poll interval), sends SIGTERM to the scanner PID recorded in `/tmp/loop-scanner.lock`, waits 5 s, then SIGKILLs if still alive.
- launchd `KeepAlive=true` on the scanner ensures automatic restart after the kill.
- Exits 0 in all cases (missing heartbeat, dead PID, fresh heartbeat) so the watchdog itself never causes a launchd storm.

- `StartInterval 300` (every 5 min); logs to `loop-scanner-watchdog.log`.

### `install.sh`
- `bootstrap_register_launchd()`: installs the watchdog plist alongside the scanner plist (idempotent).
- `bootstrap_register_cron()`: adds a `*/5` cron entry for Linux installs.

- 6 bats tests: heartbeat is created by `run_once()`, updated on each call, skipped in dry-run; watchdog correctly identifies fresh/missing/stale heartbeats.

## Test Plan
- All 6 new bats tests pass: `bats tests/scanner-heartbeat.bats`
- Manual test: `kill -STOP $SCANNER_PID` → watchdog should kill + launchd restarts within 5 min

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,25 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — fires every 5 min; kills a wedged scanner
+    # so launchd can restart it automatically.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #526, opened 2026-05-18

svv2014:
Closes #413

### `scanner/scanner.sh`
- Added `_heartbeat_write()` that writes the current epoch to `${LOOP_LOG_DIR}/scanner-heartbeat` on every tick.
- Called at the top of `run_once()` (skipped in `--dry-run` mode), so the file's content always reflects when the last tick began.

- Periodic watchdog that reads the heartbeat file and computes its age.
- Stale threshold: `2 × LOOP_SCANNER_INTERVAL` (default 600s). Configurable via `LOOP_SCANNER_INTERVAL` in `loop.env`.
- On stale: kills the scanner PID (from `/tmp/loop-scanner.lock`) then runs `launchctl kickstart` on macOS; on Linux the kill alone is sufficient since cron restarts the scanner on the next tick.
- Supports `--dry-run` for manual inspection without taking action.

- `StartInterval=300` (every 5 min) — fires independently of the scanner's own sleep loop.

### `install.sh`
- `bootstrap_register_launchd`: registers `com.user.loop-scanner-watchdog` alongside the existing scanner/reconciler plists.
- `bootstrap_register_cron`: adds `*/5 * * * *` watchdog entry for Linux.

- 6 tests covering: file creation, content format, mtime update, watchdog healthy/stale/absent scenarios.

## Test Plan
- `bats tests/scanner-heartbeat.bats` → all 6 pass
- `bats tests/scanner.bats tests/scanner-log-sighup.bats tests/scanner-stage-age.bats` → no regressions
- Manual: `kill -STOP $SCANNER_PID` → watchdog fires within 5 min and restarts via launchd

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -338,6 +338,26 @@ bootstrap_register_launchd() {
         fi
     fi
 
+    # Scanner liveness watchdog — fires every 5 min, restarts scanner if heartbeat
+    # goes stale (age > 2× poll interval). Guards against silent wedge scenarios
+    # where the scanner PID is alive but emits no events.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #536, opened 2026-05-18
[LOOP-413] scanner: liveness heartbeat + watchdog for silent-stop recovery

svv2014:
Closes #413

## Changes

**`scanner/scanner.sh`**
- Added `_write_heartbeat()` function that writes the current unix timestamp to the heartbeat file each tick
- Called `_write_heartbeat()` at the start of `run_once()` (skipped in dry-run mode so tests aren't polluted)

**`scripts/scanner-watchdog.sh`** (new)
- Reads `LOOP_SCANNER_INTERVAL` (default 300 s) to compute a stale threshold of 2× the poll interval
- Checks the mtime of `scanner-heartbeat`; if age exceeds the threshold, reads the PID from `/tmp/loop-scanner.lock` and sends `kill` to the wedged process
- launchd `KeepAlive=true` on the scanner plist auto-restarts it within the `ThrottleInterval`
- `--dry-run` flag reports stale state without killing anything

- `StartInterval 300` (every 5 min) — fires the watchdog independently of the scanner
- Logs to `loop-scanner-watchdog.log`

**`install.sh`**
- macOS: registers `com.user.loop-scanner-watchdog` via launchd alongside the existing scanner/reconciler plists
- Linux: adds `*/5 * * * *` crontab entry for `scanner-watchdog.sh` alongside existing entries

**`tests/scanner-heartbeat.bats`** (new, 8 tests)
- `_write_heartbeat` creates the file, writes an integer timestamp, updates mtime on repeated calls
- `run_once` writes heartbeat in normal mode and skips in dry-run mode
- `scanner-watchdog.sh --dry-run` reports healthy when file is fresh, WARN when stale, and never kills in dry-run

## Test Plan

- All 8 new bats tests pass
- No regressions in `scanner.bats`, `scanner-log-sighup.bats`, `scanner-stage-age.bats` (5 pre-existing failures confirmed present on `main` before this PR)
- Manual: `kill -STOP $SCANNER_PID` → watchdog fires within 5 min (2× poll interval), kills the stopped process, launchd restarts
- Dry-run smoke: `./scripts/scanner-watchdog.sh --dry-run` with a fresh heartbeat → "healthy"; with an aged file → "WARN … DRY-RUN: would kill"

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,25 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — kills a wedged (alive PID, no heartbeat)
+    # scanner so launchd KeepAlive restarts it. Runs every 5 min.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
+            -e "s|__HOME__|$HOME|g" \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #537, opened 2026-05-18
[LOOP-413] scanner: heartbeat file + stale-scanner watchdog

svv2014:
Closes #413

### scanner/scanner.sh
- Write `$(date +%s)` to `${LOOP_LOG_DIR}/scanner-heartbeat` at the start of every `run_once()` tick (skipped in `--dry-run`). This proves the scanner is actively ticking, not just alive-but-wedged.

### scanner/restart-scanner-if-stale.sh (new)
- Watchdog script intended to run every 5 minutes.
- Reads the epoch from `scanner-heartbeat`; if `now - beat >= LOOP_SCANNER_WATCHDOG_STALE_SECONDS` (default 900s = 3× poll interval), kills the scanner PID from the lock file and calls `launchctl kickstart -k` so launchd KeepAlive restarts a fresh scanner process.
- Exits immediately on non-macOS hosts (Linux uses `--once` cron ticks, which cannot wedge).
- `--dry-run` flag for safe testing.

- launchd plist that fires `restart-scanner-if-stale.sh` every 300s.

### install.sh
- Registers `com.user.loop-scanner-watchdog` in `bootstrap_register_launchd()` (idempotent, same pattern as existing services).

- 6 bats tests: scanner writes heartbeat on `--once`, epoch is recent, watchdog correctly classifies fresh/stale/absent heartbeat, threshold boundary check.

## Test Plan
- All 6 new bats tests pass locally
- `bash -n` + `shellcheck -S warning` clean on all scripts
- No personal identifiers
- Manual: `kill -STOP $SCANNER_PID` → watchdog should restart within 15 min (2–3 watchdog cycles)
- Manual: `scanner/restart-scanner-if-stale.sh --dry-run` to inspect output without side effects

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,25 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — checks heartbeat file every 5 min and kills
+    # a wedged scanner so launchd KeepAlive restarts it.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #541, opened 2026-05-18

svv2014:
Closes #413

### `scanner/scanner.sh`
- Added `HEARTBEAT_FILE="${LOOP_LOG_DIR}/scanner-heartbeat"` variable
- At the start of each `run_once()` tick, writes current timestamp + PID to the heartbeat file (`touch`-equivalent) so external monitors can check liveness
- Added log-file integrity check: if `$LOG_FILE` is not writable at tick start, attempts to reopen FDs 1+2; exits with code 1 if reopen fails so launchd/cron restarts the process

- Standalone script that runs every 5 minutes via launchd/cron
- Reads the heartbeat mtime; if older than `2 × LOOP_SCANNER_INTERVAL` (default 600 s), the scanner is considered wedged
- Finds the scanner PID from `/tmp/loop-scanner.lock` and sends SIGTERM; launchd KeepAlive then restarts it
- Supports `--dry-run` flag for observability without side effects
- Safe no-op when scanner is healthy or heartbeat file does not yet exist

- launchd plist that fires `scanner-watchdog.sh` every 300 s (StartInterval)
- Uses the same `__LOOP_ROOT__` / `__LOG_DIR__` / `__HOME__` / `__EXTRA_PATH__` substitution tokens as the other Loop plists

### `install.sh`
- `bootstrap_register_launchd`: installs and loads `com.user.loop-scanner-watchdog.plist` (idempotent, skips if already registered)
- `bootstrap_register_cron` (Linux): adds a `*/5` cron entry for `scanner-watchdog.sh` alongside the existing scanner entry

## Test Plan
- `bash -n` and `shellcheck -S warning` pass on all scripts (verified locally)
- Manual wedge simulation: `kill -STOP $SCANNER_PID` → after ~10 min the watchdog detects staleness and sends SIGTERM; launchd restarts and scanner resumes logging
- `--dry-run` mode: `scanner/scanner-watchdog.sh --dry-run` should log the stale age and "would SIGTERM" without killing anything
- Heartbeat update: run `scanner/scanner.sh --once` → verify `${LOOP_LOG_DIR}/scanner-heartbeat` is created/updated with a fresh timestamp

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,25 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — kills wedged scanner so launchd KeepAlive restarts it.
+    # Fires every 5 min; harmless no-op when scanner is healthy.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
+            -e "s|__HOME__|$HOME|g" \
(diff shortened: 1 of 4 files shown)


svv2014:

PR #544, opened 2026-05-18

svv2014:
Closes #413

### `scanner/scanner.sh`
- **Heartbeat write**: at the start of every `run_once` tick, writes the current Unix epoch to `${LOOP_LOG_DIR}/scanner-heartbeat`. Written before all other work so even a slow or partially-stuck tick keeps the watchdog satisfied.
- **Stdout integrity check**: if `LOG_FILE` exists but is no longer writable (log rotation without SIGHUP), reopens FDs 1+2 via `exec 1>>"$LOG_FILE" 2>&1`; exits on failure so launchd restarts with fresh descriptors.

### `scanner/restart-scanner-if-stale.sh` (new)
Watchdog script:
- Reads `${LOOP_LOG_DIR}/scanner-heartbeat` mtime
- Compares against `STALE_THRESHOLD = 2 × LOOP_SCANNER_INTERVAL` (default 600 s / 10 min)
- If stale: on macOS uses `launchctl kickstart -k` to restart the launchd-managed scanner; on Linux/cron sends SIGTERM to the PID in `/tmp/loop-scanner.lock`
- `--dry-run` flag logs what would happen without acting

launchd plist running `restart-scanner-if-stale.sh` every 5 minutes (`StartInterval=300`). Uses the same `__LOOP_ROOT__`/`__LOG_DIR__`/`__HOME__`/`__EXTRA_PATH__` substitution pattern as other plists.

### `install.sh`
Registers the scanner watchdog plist alongside the scanner in `bootstrap_register_launchd()`. Idempotent: skips if already present.

7 bats tests covering: heartbeat file written to `LOOP_LOG_DIR`, valid numeric epoch content, mtime updates across ticks, dry-run suppression, watchdog fresh/stale/missing heartbeat detection.

## Test Plan
- `bats tests/scanner-heartbeat.bats` — all 7 pass
- `bash -n` + `shellcheck -S warning` pass on all changed files
- Manual: `kill -STOP $SCANNER_PID` → watchdog should log restart intent within 10 min (2× 300 s poll interval)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -316,6 +316,25 @@ bootstrap_register_launchd() {
         fi
     fi
 
+    # Scanner liveness watchdog — detects a wedged scanner (alive PID, no events)
+    # and restarts it. Fires every 5 min; stale threshold is 2× poll interval.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
+            -e "s|__HOME__|$HOME|g" \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #548, opened 2026-05-18
[LOOP-413] scanner: liveness heartbeat + stale-scanner watchdog

svv2014:
Closes #413

### scanner/scanner.sh
- Added `_scanner_write_heartbeat()` — writes `<pid> <epoch> <dedup_count>` to the heartbeat file on every tick; watchdog reads mtime, not content
- Added `_scanner_check_stdout()` — at the top of each tick, checks whether the log file path is writable; if not, attempts FD reopen (same mechanism as the SIGHUP handler) and exits if that also fails so launchd restarts cleanly
- Both are called at the top of `run_once()`, skipped in `--dry-run` mode

### scripts/restart-scanner-if-stale.sh (new)
- Watchdog script: reads heartbeat file mtime; if age ≥ `LOOP_SCANNER_STALE_THRESHOLD` (default `2 × POLL_INTERVAL = 10 min`), kills the scanner PID from the lock file and asks launchd to `kickstart` the service (KeepAlive ensures restart even if kickstart fails)
- `--dry-run` flag for safe local testing

- `StartInterval 300` (every 5 min) — fires the watchdog on a separate schedule from the scanner itself

- 7 bats tests covering: heartbeat file creation, format, mtime update, dedup_count recording, watchdog OK/STALE/absent-file detection (all in `--dry-run` mode)

## Test Plan
- All 7 new tests pass: `bats tests/scanner-heartbeat.bats`
- No regressions in existing scanner tests (pre-existing failures in tests 15/16/20/21/28 are unchanged)
- Simulate wedge: `kill -STOP $SCANNER_PID` → heartbeat goes stale → watchdog fires within 10 min and restarts scanner
- Manual check: `cat ${LOOP_LOG_DIR}/scanner-heartbeat` should update on every 5-min tick

### DIFF
--- a/scanner/scanner.sh
+++ b/scanner/scanner.sh
@@ -29,6 +29,7 @@ source "$LOOP_ROOT/lib/jobs.sh"
 
 LOCK_FILE="/tmp/loop-scanner.lock"
 LOG_FILE="${LOOP_LOG_DIR}/loop-scanner.log"
+HEARTBEAT_FILE="${LOOP_LOG_DIR}/scanner-heartbeat"
 POLL_INTERVAL="${LOOP_SCANNER_INTERVAL:-300}"
 BOBA_EVENT_CLIENT="${LOOP_EVENT_CLIENT:-}"
 HANDLER_TIMEOUT="${LOOP_HANDLER_TIMEOUT:-7200}"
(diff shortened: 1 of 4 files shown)


svv2014:

PR #552, opened 2026-05-18

svv2014:
Closes #413

### scanner/scanner.sh
- **Heartbeat write**: `run_once()` now calls `touch ${LOOP_LOG_DIR}/scanner-heartbeat` at the start of every tick (skipped in `--dry-run` mode). This gives the watchdog a reliable mtime signal.
- **Stdout integrity check**: at the top of each tick, if `LOG_FILE` exists but is not writable, attempts to reopen the FD via `exec`; exits with code 1 (triggering launchd restart) if that also fails.

Standalone watchdog that fires every 5 minutes. Reads the heartbeat file mtime; if `now - mtime >= STALE_THRESHOLD` (default: `2 × LOOP_SCANNER_INTERVAL = 600s`), it:
1. Reads the scanner PID from `/tmp/loop-scanner.lock`
2. Sends SIGTERM, waits 5 s, then SIGKILL if still alive
3. Removes the stale lock file
4. On macOS, sends `launchctl kickstart` so the scanner restarts immediately rather than waiting for ThrottleInterval

Configurable via `LOOP_SCANNER_WATCHDOG_THRESHOLD` and `LOOP_SCANNER_INTERVAL`.

macOS launchd plist for the watchdog. Fires every 300 s (StartInterval), no RunAtLoad (avoids spurious restart before the first tick).

### install.sh
- **macOS**: registers `com.user.loop-scanner-watchdog` plist in `bootstrap_register_launchd`
- **Linux**: adds `*/5 * * * *` crontab entry in `bootstrap_register_cron`

Both are idempotent (skip if already registered).

Three bats tests verifying:
- `run_once` creates the heartbeat file
- `run_once` updates the mtime on subsequent ticks
- heartbeat is skipped in `--dry-run` mode

## Test Plan
- `bats tests/scanner-heartbeat.bats` — all 3 pass
- `bash -n` + `shellcheck -S warning` — no new errors
- Personal-identifier grep — clean
- Pre-existing scanner.bats failures (tests 15, 16, 20, 21) are unrelated to this PR (also fail on main)
- Manual: `kill -STOP $SCANNER_PID` → heartbeat goes stale → watchdog kills + launchd restarts within 10 min

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,26 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — fires every 5 min and restarts the scanner
+    # if its heartbeat file has not been updated within 2× the poll interval.
+    # Guards against silent wedge (alive PID, no log activity).
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #553, opened 2026-05-18

svv2014:
Closes #413

## Changes

**`scanner/scanner.sh`**
- `run_once()` now writes `${LOOP_LOG_DIR}/scanner-heartbeat` (unix epoch) at the top of every tick. Skip in `--dry-run` mode.
- Added log file integrity guard: if `$LOG_FILE` exists but is not writable (e.g. permissions changed post-rotation without SIGHUP), exit immediately so launchd restarts the scanner with fresh file descriptors.

- Standalone script that reads the heartbeat file mtime and compares against `LOOP_SCANNER_STALE_THRESHOLD` (default: `2 × LOOP_SCANNER_INTERVAL = 600s`).
- If stale: sends SIGTERM to the scanner PID from the lock file (SIGKILL after 2s if still alive), then exits — launchd KeepAlive restarts the scanner automatically.
- If no heartbeat file: exits silently (scanner may be in its first tick).
- Supports `--dry-run` for manual inspection.

- Runs `scanner-watchdog.sh` every 5 min (`StartInterval=300`).
- Shares `__LOOP_ROOT__`/`__LOG_DIR__`/`__HOME__`/`__EXTRA_PATH__` substitution pattern with all other launchd templates.

**`install.sh`**
- `bootstrap_register_launchd`: registers the scanner watchdog plist alongside scanner/reconciler/pr-watchdog.
- `bootstrap_register_cron`: adds `*/5` cron entry for the watchdog on Linux.

- `run_once` writes heartbeat file with a numeric epoch timestamp.
- `run_once` updates the heartbeat on successive calls.
- `--dry-run` mode skips the heartbeat write.
- Watchdog exits OK when heartbeat is fresh.
- Watchdog exits OK when heartbeat file does not exist yet.
- Watchdog reports stale heartbeat + `DRY-RUN` output when file is back-dated.

## Test Plan
- All 6 new bats tests pass (`bats tests/scanner-heartbeat.bats`).
- `bash -n` and `shellcheck -S warning` clean on all scripts.
- No personal identifiers in committed files.
- Manual: `kill -STOP $SCANNER_PID` for >600s → watchdog fires and kills the PID; launchd restarts scanner within seconds.
- Manual: `ls -la ~/.loop/logs/scanner-heartbeat` after a tick — mtime updates each poll cycle.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -316,6 +316,25 @@ bootstrap_register_launchd() {
         fi
     fi
 
+    # Scanner liveness watchdog — kills a wedged scanner (alive PID, no heartbeat
+    # updates) so launchd KeepAlive restarts it. Fires every 5 min.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
+            -e "s|__HOME__|$HOME|g" \
+            -e "s|__EXTRA_PATH__|$extra_path|g" \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #556, opened 2026-05-19

svv2014:
Closes #413

## Changes

**`scanner/scanner.sh`**
- At the start of every `run_once()` tick, `touch ${LOOP_LOG_DIR}/scanner-heartbeat` so the watchdog has a reliable liveness signal.
- Added stdout integrity check at tick start: if `LOG_FILE` is not writable (e.g. inode deleted after log rotation), reopen the FD or exit so launchd auto-restarts.
- Both changes are no-ops in `--dry-run` mode.

- Reads heartbeat file mtime; if age exceeds `LOOP_WATCHDOG_STALE_THRESHOLD` (default: `2 × LOOP_SCANNER_INTERVAL = 600s`), kills the scanner PID from the lock file and triggers restart.
- On macOS: `launchctl kickstart` the scanner plist (KeepAlive handles the restart; kickstart is belt-and-braces).
- On Linux: spawns `scanner.sh --once` in the background.

- Fires every 5 minutes (`StartInterval=300`); `RunAtLoad=false` (watchdog should not fire before the scanner has had a chance to write its first heartbeat).

**`install.sh`**
- `bootstrap_register_launchd`: installs watchdog plist alongside scanner plist.
- `bootstrap_register_cron`: adds `*/5 * * * *` watchdog cron entry for Linux.

- `run_once` creates and updates heartbeat file; skips in dry-run.
- Watchdog reports healthy when heartbeat is fresh.
- Watchdog reports stale (with dry-run restart) when heartbeat is absent or old.

## Test Plan
- All new bats tests pass (`bats tests/scanner-heartbeat.bats` → 6/6 ok).
- `bash -n` and `shellcheck -S warning` pass on all shell scripts.
- Pre-existing `scanner.bats` failures (5 tests) are unrelated to this PR and were present before this change.
- Manual: simulate wedge with `kill -STOP $SCANNER_PID`; watchdog should restart within 10 minutes (2×poll).

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -338,6 +338,25 @@ bootstrap_register_launchd() {
         fi
     fi
 
+    # Scanner liveness watchdog — fires every 5 min, checks heartbeat file mtime,
+    # kills and restarts the scanner if it has been silent for 2×poll interval.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
+            -e "s|__HOME__|$HOME|g" \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #558, opened 2026-05-19
[LOOP-413] scanner: heartbeat file + watchdog for silent-hang detection

svv2014:
Closes #413

### `scanner/scanner.sh`
- **Heartbeat file**: `run_once()` now calls `touch ${LOOP_LOG_DIR}/scanner-heartbeat` at the top of every tick (skipped in `--dry-run`). The mtime of this file is the canonical liveness signal.
- **Stdout integrity check**: if `loop-scanner.log` exists but is not writable (e.g. stale inode after log rotation was not caught by SIGHUP), the scanner exits immediately so launchd KeepAlive triggers a clean restart.

Standalone watchdog script designed to run every 5 minutes via launchd. Reads `scanner-heartbeat` mtime; if age ≥ `2 × LOOP_SCANNER_INTERVAL` (default 10 min):
1. Kills the wedged scanner PID (from the lock file).
2. On macOS calls `launchctl kickstart` as belt-and-braces (KeepAlive auto-restarts regardless).
3. On Linux, clearing the lock is sufficient for the next cron invocation.
Supports `--dry-run` for operator inspection without side-effects.

LaunchAgent template that fires `scanner-watchdog.sh` every 300 s (5 min). Uses the same `__LOOP_ROOT__` / `__LOG_DIR__` / `__HOME__` / `__EXTRA_PATH__` placeholders as the existing scanner plist.

### `install.sh`
`bootstrap_register_launchd()` now installs and loads `com.user.loop-scanner-watchdog` alongside the scanner and reconciler plists (idempotent — skips if plist already present).

Three bats tests:
- Heartbeat file is created on first tick.
- Heartbeat mtime is updated on subsequent ticks.
- Heartbeat is NOT written in `--dry-run` mode.

## Test Plan
- All bats tests pass: `bats tests/scanner-heartbeat.bats`
- CI: `bash -n` + `shellcheck -S warning` + workflow lint + no-personal-identifiers — all clean
- Manual: simulate wedge with `kill -STOP $SCANNER_PID`, observe watchdog log after ~10 min; scanner should be killed and restarted within 5 min of the watchdog firing

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -316,6 +316,26 @@ bootstrap_register_launchd() {
         fi
     fi
 
+    # Scanner liveness watchdog — detects silent scanner hangs by checking
+    # heartbeat file mtime and kills+restarts the scanner if stale.
+    # Fires every 5 min; stale threshold is 2 × poll interval (default 10 min).
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
+            -e "s|__HOME__|$HOME|g" \
+            -e "s|__EXTRA_PATH__|$extra_path|g" \
+            "$template_dir/com.user.loop-scanner-watchdog.plist.template" > "$scanner_watchdog_plist"
+        if launchctl load "$scanner_watchdog_plist" 2>/dev/null; then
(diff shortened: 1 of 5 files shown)


svv2014 (commented):
[loop:review] Heartbeat file, watchdog script, launchd plist, install.sh integration, and bats tests all look solid — lock file path in the watchdog matches the scanner exactly (`/tmp/loop-scanner.lock`), stat cross-platform fallback is correct, and `--dry-run` coverage is present. CI is clean (lint + closes-ref pass). AC item #2 (`scanner_tick` event to loop-monitor) is unimplemented, but that depends on a monitoring dashboard that does not yet exist and the issue itself offered an "alternative simpler implementation" which this PR follows. Code is approved on the strength of the core hang-detection mechanism, which fully addresses the stated problem. Advancing to QA.

svv2014:

PR #560, opened 2026-05-19

svv2014:
Closes #413

### `scanner/scanner.sh`
- Added `HEARTBEAT_FILE` variable (`${LOOP_LOG_DIR}/scanner-heartbeat`).
- `run_once()` now calls `touch "$HEARTBEAT_FILE"` at the start of every tick (skipped in dry-run mode), giving any external watchdog a reliable mtime-based liveness signal.

- Reads the heartbeat file's mtime. If it is older than `LOOP_SCANNER_STALE_THRESHOLD` (default 900 s = 15 min, 3× the default 5-min poll interval), reads the scanner PID from `/tmp/loop-scanner.lock` and kills it so launchd/systemd's `KeepAlive`/`Restart` policy triggers an automatic restart.
- Supports `--dry-run` to report status without killing anything.
- Exits cleanly when the heartbeat file has never been written (first start) or the PID is already gone.

- launchd plist that runs `restart-scanner-if-stale.sh` every 5 minutes.
- Designed to be installed alongside the existing scanner plist.

## Test Plan

- `bats tests/scanner-heartbeat.bats` — 7 tests:
  - `run_once` writes heartbeat on every tick (non-dry-run).
  - `run_once` updates mtime on repeated ticks.
  - Dry-run does NOT write heartbeat.
  - Watchdog exits cleanly when heartbeat absent.
  - Watchdog reports healthy for a fresh heartbeat.
  - Watchdog detects stale heartbeat and reports "wedged".
  - Dry-run does not kill PID even when stale.
- Manual: `kill -STOP $SCANNER_PID` → watchdog kills it within 15 min, launchd restarts.

### DIFF
--- a/scanner/scanner.sh
+++ b/scanner/scanner.sh
@@ -29,6 +29,7 @@ source "$LOOP_ROOT/lib/jobs.sh"
 
 LOCK_FILE="/tmp/loop-scanner.lock"
 LOG_FILE="${LOOP_LOG_DIR}/loop-scanner.log"
+HEARTBEAT_FILE="${LOOP_LOG_DIR}/scanner-heartbeat"
 POLL_INTERVAL="${LOOP_SCANNER_INTERVAL:-300}"
 BOBA_EVENT_CLIENT="${LOOP_EVENT_CLIENT:-}"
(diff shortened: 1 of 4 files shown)


svv2014:

PR #564, opened 2026-05-19

svv2014:
Closes #413

### scanner/scanner.sh
- `run_once()` now writes `<epoch> dedup_count=<N>` to `${LOOP_LOG_DIR}/scanner-heartbeat` at the start of every tick (before any blocking I/O), making wedge detection possible via file mtime.
- Added a stdout integrity check: if `LOG_FILE` exists but is no longer writable, the process exits so launchd KeepAlive can restart it with fresh FDs.

- Reads the heartbeat file mtime; if age exceeds `LOOP_SCANNER_INTERVAL × LOOP_WATCHDOG_STALE_FACTOR` (default: 2 × 300s = 10 min), kills the scanner PID so launchd restarts it.
- Configurable via `LOOP_WATCHDOG_STALE_FACTOR` env var.

- Runs `scanner-watchdog.sh` every 5 min (StartInterval 300) via launchd on macOS.

### install.sh
- `bootstrap_register_launchd()` now installs the scanner watchdog plist alongside the existing scanner plist.
- `bootstrap_register_cron()` now adds a `*/5` cron entry for the watchdog on Linux.

- 8 bats tests covering: heartbeat file creation, timestamp correctness, dedup_count field, mtime update on second call, dry-run skips, and all three watchdog states (no file, fresh, stale).

## Test Plan
- All 8 new bats tests pass: `bats tests/scanner-heartbeat.bats`
- No regressions in `tests/scanner.bats` (4 pre-existing failures unrelated to this change)
- `bash -n` and `shellcheck -S warning` both clean
- Manual simulation: `kill -STOP $SCANNER_PID` → heartbeat goes stale → watchdog detects after ≤10 min → scanner restarts

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,25 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — fires every 5 min, kills a wedged scanner
+    # (alive PID, stale heartbeat) so launchd KeepAlive restarts it cleanly.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #569, opened 2026-05-19
[LOOP-413] scanner: liveness heartbeat + stale-restart watchdog

svv2014:
Closes #413

## Changes

**`scanner/scanner.sh`**
- Added `HEARTBEAT_FILE="${LOOP_LOG_DIR}/scanner-heartbeat"` variable.
- Added `_scanner_write_heartbeat()`: writes `YYYY-MM-DDTHH:MM:SS pid=$$` to the heartbeat file at the start of every tick. No-op in dry-run. Write errors are silently ignored so a transient failure never wedges the tick.
- Added `_scanner_check_stdout()`: checks `LOG_FILE` writability before each tick; attempts one FD reopen via the existing SIGHUP handler, then exits (triggering launchd restart) if still unwritable.
- Modified `run_once()`: calls both helpers and logs `heartbeat=<ts> dedup_entries=<n>` in the tick-start line, providing in-log telemetry matching the `scanner_tick` requirement.

**`scanner/restart-scanner-if-stale.sh`** (new, executable)
- Reads `HEARTBEAT_FILE` mtime; skips if file is absent (scanner may be starting up).
- If age ≥ `LOOP_SCANNER_STALE_THRESHOLD` (env, default `2 × LOOP_SCANNER_INTERVAL = 600s`), kills the PID from the lock file and lets launchd/cron restart scanner.sh.
- All kill/stat calls are fail-safe; the script always exits 0 so launchd does not throttle it.
- Respects `LOOP_SCANNER_LOCK` env var for testability.

- Fires `restart-scanner-if-stale.sh` every 5 minutes (`StartInterval: 300`).
- Uses the same `__LOOP_ROOT__` / `__LOG_DIR__` / `__HOME__` / `__EXTRA_PATH__` placeholder convention as the other launchd templates.

- `_scanner_write_heartbeat`: creates file, contains ISO timestamp, updates mtime on successive calls, is a no-op in dry-run mode.
- `restart-scanner-if-stale.sh`: exits 0 on fresh heartbeat; exits 0 with "absent" message when file missing; kills a live PID and reports STALE when heartbeat is backdated past threshold.

## Test Plan
- `bash -n lib/*.sh scripts/*.sh scanner/*.sh install.sh` — all pass.
- `shellcheck -S warning lib/*.sh scripts/*.sh scanner/*.sh install.sh` — no warnings.
- `bats tests/scanner-heartbeat.bats` — all 7 tests pass.
- Manual: run scanner, confirm `${LOOP_LOG_DIR}/scanner-heartbeat` is updated every tick.
- Manual wedge simulation: `kill -STOP $SCANNER_PID` → watchdog fires within 5 min → logs "STALE" → kills PID → launchd restarts scanner.

### DIFF
--- a/scanner/restart-scanner-if-stale.sh
+++ b/scanner/restart-scanner-if-stale.sh
@@ -1,0 +1,76 @@
+#!/usr/bin/env bash
+# restart-scanner-if-stale.sh — scanner liveness watchdog.
+#
+# Reads ${LOOP_LOG_DIR}/scanner-heartbeat. If the file is absent or its mtime
+# is older than LOOP_SCANNER_STALE_THRESHOLD seconds (default: 2 × POLL_INTERVAL
+# = 600s), the scanner is considered wedged. Kills the PID recorded in
+# LOOP_SCANNER_LOCK (default: /tmp/loop-scanner.lock); launchd (KeepAlive=true)
+# or cron will restart scanner.sh automatically.
+#
+# Intended to run every 5 minutes via launchd StartInterval or cron.
+#
+# macOS install — see templates/launchd/com.user.loop-scanner-watchdog.plist.template
+# Linux crontab:
+#   */5 * * * * /path/to/loop/scanner/restart-scanner-if-stale.sh
+
+set -euo pipefail
+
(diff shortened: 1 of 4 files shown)


svv2014:

PR #574, opened 2026-05-19
[LOOP-413] scanner: heartbeat file + watchdog to detect and restart silent stops

svv2014:
Closes #413

### `scanner/scanner.sh`
- At the start of every `run_once()` tick, writes `date +%s` (unix timestamp) to `${LOOP_LOG_DIR}/scanner-heartbeat`. Skipped in `--dry-run` mode. This gives an external watchdog a ground-truth signal that the scanner is still alive and iterating.

- Reads the heartbeat file mtime. If it is older than `LOOP_SCANNER_STALE_SECONDS` (default 900 s = 15 min = 3× the default poll interval), considers the scanner silently wedged.
- Kills the scanner PID (from `/tmp/loop-scanner.lock`) so launchd's `KeepAlive=true` restarts it automatically on macOS.
- On macOS also calls `launchctl kickstart` for an immediate restart rather than waiting for launchd's throttle window.
- Exits cleanly if the heartbeat file is absent (scanner never started) or mtime is fresh.

- `StartInterval=300` (every 5 min).

### `install.sh`
- `bootstrap_register_launchd`: registers `com.user.loop-scanner-watchdog` plist (idempotent, skips if already loaded).
- `bootstrap_register_cron`: adds a `*/5 * * * *` cron entry for the watchdog on Linux.

- 4 tests: file is created, content is a unix timestamp, file is absent in dry-run, mtime advances across ticks.

## Test Plan
- `bats tests/scanner-heartbeat.bats` — all 4 pass
- `bats tests/scanner.bats` — all 45 existing tests pass
- `bash -n lib/*.sh scripts/*.sh scanner/*.sh install.sh` — no errors
- `shellcheck -S warning lib/*.sh scripts/*.sh scanner/*.sh install.sh` — no warnings
- Manual: `kill -STOP $SCANNER_PID` → watchdog should restart scanner within 15 min (or sooner via `launchctl kickstart`)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,25 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — fires every 5 min, kills stale scanner so
+    # launchd's KeepAlive restarts it automatically.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #577, opened 2026-05-19
[LOOP-413] scanner: heartbeat file + stdout integrity check + watchdog

svv2014:
Closes #413

### scanner/scanner.sh
- Added `_scanner_write_heartbeat()` — writes current timestamp to `${LOOP_LOG_DIR}/scanner-heartbeat` on every tick (skipped in dry-run mode). Provides a liveness signal that a separate watchdog can monitor.
- Added `_scanner_check_stdout()` — called at the top of every tick; if `LOG_FILE` was rotated or deleted, it reopens FDs 1+2 against the current path. If reopen fails, exits immediately so launchd/cron restarts the scanner cleanly. Belt-and-suspenders complement to the existing SIGHUP handler.
- Updated `run_once()` to call both functions and emit a structured tick log line: `=== scan tick start === ts=<epoch> dedup_count=<n>` so silent-stop becomes visible in the log.
- Added `HEARTBEAT_FILE` variable derived from `LOOP_LOG_DIR`.

Standalone script run every 5 minutes via launchd/cron. Reads heartbeat file mtime; if older than `LOOP_SCANNER_WATCHDOG_STALE` seconds (default: 2 × poll interval = 600 s), kills the scanner PID from the lock file. On macOS with `LOOP_LAUNCHD_LABEL` set, also calls `launchctl kickstart` for immediate restart. Supports `--dry-run`.

launchd plist template for the watchdog. Fires every 300 s (`StartInterval`). Sets `LOOP_LAUNCHD_LABEL=com.user.loop-scanner` so the kickstart path is wired by default.

8 bats tests covering:
- Heartbeat file is created and updated on each call
- Content is a non-empty timestamp
- No-op in dry-run mode
- `run_once()` writes heartbeat on every tick
- `_scanner_check_stdout()` no-ops when file is writable or LOG_FILE is unset

## Test Plan
- All 8 new bats tests pass: `bats tests/scanner-heartbeat.bats`
- Existing scanner tests show no regressions (pre-existing failures on main are unchanged)
- Syntax check: `bash -n scanner/scanner-watchdog.sh scanner/scanner.sh` ✓
- shellcheck: clean ✓
- Simulate wedge: `kill -STOP $SCANNER_PID` → watchdog kills and restarts within 10 min (2× poll interval)
- Dry-run: `scanner/scanner-watchdog.sh --dry-run` logs state without restarting

### DIFF
--- a/scanner/scanner-watchdog.sh
+++ b/scanner/scanner-watchdog.sh
@@ -1,0 +1,97 @@
+#!/usr/bin/env bash
+# scanner-watchdog.sh — restart the scanner if it stops writing its heartbeat.
+#
+# Reads ${LOOP_LOG_DIR}/scanner-heartbeat. If the file is absent or its mtime
+# is older than STALE_THRESHOLD_SECONDS (default: 2 × poll interval = 600s),
+# the scanner is considered wedged.  The watchdog kills the PID stored in
+# /tmp/loop-scanner.lock and relies on launchd (macOS) or cron (Linux) to
+# restart the scanner binary.  On macOS it additionally calls
+# `launchctl kickstart` if LOOP_LAUNCHD_LABEL is set.
+#
+# Usage:
+#   scanner-watchdog.sh            # single check (default; run from launchd / cron)
+#   scanner-watchdog.sh --dry-run  # report state without restarting
+
+set -euo pipefail
+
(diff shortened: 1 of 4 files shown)


svv2014:

PR #579, opened 2026-05-19
[LOOP-413] scanner: heartbeat file + watchdog auto-restart for silent wedge

svv2014:
Closes #413

## Changes

**`scanner/scanner.sh`** — write `${LOOP_LOG_DIR}/scanner-heartbeat` (via `touch`) at the top of every `run_once()` tick. This is the liveness signal that the watchdog consumes.

**`scripts/scanner-watchdog.sh`** — new script, runs every 5 minutes. Reads the heartbeat file mtime; if age exceeds `LOOP_SCANNER_WATCHDOG_THRESHOLD` (default `2 × POLL_INTERVAL` = 600 s), kills the PID from `/tmp/loop-scanner.lock` and lets launchd/cron restart the scanner automatically. `--dry-run` flag reports but does not kill. Respects the same `LOOP_SCANNER_INTERVAL` env var as the scanner so threshold auto-scales if the poll interval is changed.

**`templates/launchd/com.user.loop-scanner-watchdog.plist.template`** — launchd plist, `StartInterval=300` (every 5 min). Uses the same `__TOKEN__` substitution pattern as other Loop plists.

**`install.sh`** — registers `com.user.loop-scanner-watchdog` via launchd on macOS and adds a `*/5 * * * *` cron entry on Linux. Both registrations are idempotent (skip if already present).

**`tests/scanner-heartbeat.bats`** — 5 bats tests:
1. `run_once` creates the heartbeat file on first tick
2. `run_once` updates the heartbeat mtime on each tick
3. Watchdog exits 0 (healthy) when heartbeat is fresh
4. Watchdog kills the wedged scanner PID when heartbeat is stale
5. Watchdog `--dry-run` reports stale but does not kill

## Test Plan

- All 5 bats tests in `tests/scanner-heartbeat.bats` pass locally.
- `bash -n` syntax check: all scripts pass.
- `shellcheck -S warning`: clean.
- Personal-identifiers grep: no hits.
- Manual: simulate wedge with `kill -STOP $SCANNER_PID` → watchdog should kill and launchd restarts within the next 5-min tick.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -338,6 +338,25 @@ bootstrap_register_launchd() {
         fi
     fi
 
+    # Scanner liveness watchdog — kills and lets launchd restart a wedged scanner.
+    # Runs every 5 min; harmless when the scanner is healthy.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
+            -e "s|__HOME__|$HOME|g" \
+            -e "s|__EXTRA_PATH__|$extra_path|g" \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #581, opened 2026-05-19
[LOOP-413] scanner: liveness heartbeat + stale-scanner watchdog

svv2014:
Closes #413

## Changes

**`scanner/scanner.sh`**
- Added `HEARTBEAT_FILE="${LOOP_LOG_DIR}/scanner-heartbeat"` variable
- At the start of every `run_once()` tick: touch the heartbeat file so the external watchdog can detect a silently wedged scanner (alive PID, no tick activity)
- Added log-file writability check at tick start: if `LOG_FILE` exists but is not writable (filesystem read-only, disk full, post-rotation inode mismatch without SIGHUP), exit 1 so launchd restarts us cleanly

**`scanner/restart-scanner-if-stale.sh`** (new)
- Watchdog script to be run every 5 minutes via launchd or cron
- Reads heartbeat file mtime; if age ≥ `LOOP_SCANNER_STALE_THRESHOLD` (default 900 s = 3 × poll interval), kills the wedged scanner PID — launchd auto-restarts via `KeepAlive=true`
- Falls back to `launchctl kickstart` when PID is no longer alive
- Supports `--dry-run` for safe testing

- launchd plist template for the watchdog; fires every 5 minutes (`StartInterval=300`)

- 6 tests: heartbeat file created on first tick, mtime advances between ticks, path uses LOOP_LOG_DIR; watchdog exits cleanly when heartbeat is fresh, when file is missing (first start), and logs restart intent when stale

## Test Plan

- All 6 new bats tests pass
- `bash -n` and `shellcheck -S warning` pass on all modified/new shell scripts
- Manual: simulate wedge with `kill -STOP $SCANNER_PID`; watchdog should restart within 15 min (3 missed ticks)

### DIFF
--- a/scanner/restart-scanner-if-stale.sh
+++ b/scanner/restart-scanner-if-stale.sh
@@ -1,0 +1,84 @@
+#!/usr/bin/env bash
+# restart-scanner-if-stale.sh — watchdog: restart the scanner if its heartbeat
+# file has not been updated recently, which indicates the scanner is silently
+# wedged (alive PID, sleep loop intact, but no tick activity).
+#
+# Intended to run every 5 minutes via launchd (macOS) or cron (Linux). Because
+# the scanner plist uses KeepAlive=true, killing the wedged PID is sufficient —
+# launchd will restart the scanner automatically within ThrottleInterval seconds.
+#
+# Configuration (via loop.env or environment):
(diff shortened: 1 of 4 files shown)


svv2014:

PR #582, opened 2026-05-19

svv2014:
Closes #413

## Changes

**`scanner/scanner.sh`** — write `${LOOP_LOG_DIR}/scanner-heartbeat` on every tick via `touch`. Skipped in `--dry-run` mode. One line added to `run_once`.

**`scanner/scanner-watchdog.sh`** (new) — runs every 5 min via launchd/cron. Reads the heartbeat file mtime; if `now - mtime > 2 × POLL_INTERVAL` (default 10 min), considers the scanner wedged, kills its PID from `/tmp/loop-scanner.lock`, and removes the stale lock so launchd's `KeepAlive` restart can acquire it immediately. Includes `--dry-run` flag.

**`templates/launchd/com.user.loop-scanner-watchdog.plist.template`** (new) — `StartInterval=300`, no `KeepAlive` (one-shot per interval), uses standard `__LOOP_ROOT__`/`__LOG_DIR__`/`__HOME__`/`__EXTRA_PATH__` substitution tokens.

**`install.sh`** — registers the watchdog in both `bootstrap_register_launchd` (macOS) and `bootstrap_register_cron` (Linux). Idempotent — skips if already registered.

**`tests/scanner-heartbeat.bats`** (new) — 3 bats tests:
- heartbeat file created on first tick
- heartbeat mtime updated on subsequent ticks
- heartbeat not written in dry-run mode

## Test Plan

- Bats: `bats tests/scanner-heartbeat.bats` → 3/3 pass
- Manual: `kill -STOP $SCANNER_PID` → watchdog fires within 10 min, kills the stopped PID, launchd restarts scanner
- Manual: `scanner-watchdog.sh --dry-run` → prints verdict without killing
- Syntax: `bash -n` + `shellcheck -S warning` → clean

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,25 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — detects wedged scanner and kills it so
+    # launchd restarts it. Runs every 5 min on its own cadence.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
+            -e "s|__HOME__|$HOME|g" \
+            -e "s|__EXTRA_PATH__|$extra_path|g" \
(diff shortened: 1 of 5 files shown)


svv2014 (commented):
[loop:review] Heartbeat + watchdog implementation is clean and solid: `touch` at top of `run_once()` is correctly placed so a wedged tick prevents future updates, the `stat` cross-platform fallback (`-f%m` / `-c%Y`) is correct, and the lock-file race is handled conservatively (re-check PID before delete). `LOOP_SCANNER_INTERVAL` is consistent with `scanner.sh` (both default 300s). Bats tests cover the three key cases (create, update, dry-run skip). Issue #413 acceptance items 2 (tick-counter event) and 3 (stdout fd integrity check) are not addressed here — they are explicitly offered as optional in the issue body and should be tracked separately if desired. VERDICT: APPROVE — advancing to QA.

svv2014:

PR #583, opened 2026-05-19

svv2014:
Closes #413

## Changes

**`scanner/scanner.sh`** — heartbeat touch on every tick
- Added `touch "${LOOP_LOG_DIR}/scanner-heartbeat"` at the start of `run_once()`.
- Skipped in `--dry-run` mode to keep test environments clean.
- The watchdog reads this file's mtime to determine scanner liveness.

- If age > `LOOP_SCANNER_INTERVAL * 2` (default 600s), reads the scanner PID from `/tmp/loop-scanner.lock` and sends SIGTERM (falls back to SIGKILL).
- launchd KeepAlive=true on the scanner restarts it within seconds.
- No-ops gracefully: no heartbeat file (startup race), empty/dead lock PID, `--dry-run` flag.

- `StartInterval=300` — fires every 5 minutes regardless of scanner state.
- `RunAtLoad=false` — avoids false trigger before the first scanner tick.

**`install.sh`**
- `bootstrap_register_launchd()`: installs and loads the watchdog plist (macOS).
- `bootstrap_register_cron()`: adds `*/5 * * * *` cron entry (Linux).
- Both are idempotent (skip if already registered).

**`tests/scanner.bats`**
- `run_once: heartbeat file is created/updated on each tick` — verifies the file exists after a `run_once()` call.
- `run_once: heartbeat mtime advances between ticks` — backdates the file between ticks and confirms mtime is refreshed.

## Test Plan
- CI: `bash -n` and `shellcheck -S warning` pass on all changed scripts.
- New bats tests pass locally.
- Manual: `kill -STOP $SCANNER_PID` → watchdog should kill and launchd restarts within 10 min.

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,25 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — kills a wedged scanner so launchd restarts it.
+    # Runs every 5 min; acts only when the heartbeat file is older than 2 * poll interval.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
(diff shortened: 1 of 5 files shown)


svv2014:

PR #584, opened 2026-05-19

svv2014:
Closes #413

### scanner/scanner.sh
- `run_once()` writes the current Unix timestamp to the heartbeat file at the top of each tick (skipped in `--dry-run` mode)

- Reads the heartbeat file mtime; if absent or older than `LOOP_SCANNER_INTERVAL × LOOP_WATCHDOG_MULTIPLIER` (default: 2 × 300s = 10 min), kills the wedged scanner PID from the lock file and removes the lock so launchd/cron restarts it cleanly
- Falls back gracefully when the heartbeat file doesn't exist yet (checks for a live lock file before acting)
- Supports `--dry-run` mode for safe testing

- Runs `scanner-watchdog.sh` every 5 minutes via launchd on macOS

### install.sh
- `bootstrap_register_launchd()`: registers `com.user.loop-scanner-watchdog` plist (macOS)
- `bootstrap_register_cron()`: adds `scanner-watchdog.sh` cron entry every 5 min (Linux)

- 4 bats tests: heartbeat file is created on first tick, contains a valid Unix timestamp, has its mtime updated on each subsequent tick, and is NOT written in dry-run mode

## Test Plan
- All new bats tests pass: `bats tests/scanner-heartbeat.bats` → 4/4 ok
- No personal identifiers detected
- Manual: simulate wedge with `kill -STOP $SCANNER_PID`; watchdog should kill and remove lock within one 5-min tick, letting launchd restart the scanner

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,26 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — kills a wedged scanner so launchd restarts it.
+    # Runs every 5 min; acts only when the scanner heartbeat is older than
+    # 2 × LOOP_SCANNER_INTERVAL (default: 10 min).
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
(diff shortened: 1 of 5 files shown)


svv2014:

PR #585, opened 2026-05-19

svv2014:
Closes #413

### `scanner/scanner.sh`
- Added `_scanner_heartbeat()`: writes the current timestamp to `${LOOP_LOG_DIR}/scanner-heartbeat` at the start of every tick. Skipped in `--dry-run` mode. This file is the signal the watchdog reads to determine liveness.
- Added `_scanner_check_log_fd()`: exits with code 1 if the log file exists but is not writable (broken fd state). Missing file is allowed — normal on first startup. Exiting triggers launchd/cron to restart and reopen a fresh fd.
- `run_once()` now calls both helpers before the tick body.

- Reads the heartbeat file mtime; if age > `LOOP_SCANNER_WATCHDOG_THRESHOLD` (default 2× poll interval = 10 min), kills the scanner PID and uses `launchctl kickstart` on macOS (or removes the lock file on Linux) so the scanner restarts.
- Designed to run every 5 minutes via launchd or cron.
- No-ops cleanly with `--dry-run`.

- `StartInterval=300` (5 min), logs to `loop-scanner-watchdog.log`.

### `install.sh`
- `bootstrap_register_launchd`: registers `com.user.loop-scanner-watchdog`.
- `bootstrap_register_cron`: adds `*/5` cron entry for the watchdog.

- Verifies heartbeat file is created by `run_once`.
- Verifies heartbeat file contains a timestamp.
- Verifies mtime is updated on successive calls.
- Verifies heartbeat is suppressed in dry-run mode.
- Directly tests `_scanner_heartbeat`.

## Test Plan
- All 5 new bats tests pass: `bats tests/scanner-heartbeat.bats`
- No regressions in existing scanner tests (pre-existing failures on main are unchanged)
- Syntax: `bash -n` + `shellcheck -S warning` on all scripts: clean
- Personal-identifiers grep: clean
- Manual: `kill -STOP $SCANNER_PID` → watchdog fires within 10 min and restarts

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -338,6 +338,23 @@ bootstrap_register_launchd() {
         fi
     fi
 
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
+            -e "s|__HOME__|$HOME|g" \
+            -e "s|__EXTRA_PATH__|$extra_path|g" \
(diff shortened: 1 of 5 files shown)


svv2014:

PR #586, opened 2026-05-19

svv2014:
Closes #413

### `scanner/scanner.sh`
- Added `_write_heartbeat()` — writes `date +%s` to the heartbeat file at the start of every tick; no-op in `--dry-run` mode
- `run_once()` now calls `_write_heartbeat` as its first action

- Standalone watchdog script that reads `scanner-heartbeat` mtime
- If age > `LOOP_WATCHDOG_STALE_MULT × LOOP_SCANNER_INTERVAL` (default 2 × 300 = 600 s), sends SIGTERM then SIGKILL (with `LOOP_WATCHDOG_KILL_GRACE` grace period) to the scanner PID from the lock file
- Launchd `KeepAlive=true` on the scanner means it restarts automatically after the kill
- Supports `--dry-run` for safe testing

- LaunchAgent that fires the watchdog every 5 min (`StartInterval=300`)
- Follows same `__LOOP_ROOT__` / `__LOG_DIR__` substitution pattern as existing templates

### `install.sh`
- `bootstrap_register_launchd`: registers `com.user.loop-scanner-watchdog` plist (idempotent, skips if already present)
- `bootstrap_register_cron`: adds `*/5 * * * * scanner-watchdog.sh` cron entry for Linux (idempotent)

## Test Plan

- `tests/scanner-heartbeat.bats` (7 tests — all pass):
  - `_write_heartbeat` creates the file and writes a recent epoch timestamp
  - `_write_heartbeat` updates mtime on repeated calls
  - `_write_heartbeat` is a no-op in `--dry-run` mode
  - Watchdog exits 0 with "healthy" when heartbeat is fresh
  - Watchdog logs stale warning when heartbeat is old and no lock present
  - Watchdog logs DRY-RUN / would-kill when heartbeat stale and PID is alive
- All CI checks pass: `bash -n`, `shellcheck -S warning`, workflow lint, no personal identifiers

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -338,6 +338,25 @@ bootstrap_register_launchd() {
         fi
     fi
 
+    # Scanner liveness watchdog — restarts a wedged scanner that has a live PID
+    # but emits no events. Fires every 5 min; kills + lets launchd restart.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
+            -e "s|__LOG_DIR__|$log_dir|g" \
+            -e "s|__HOME__|$HOME|g" \
+            -e "s|__EXTRA_PATH__|$extra_path|g" \
+            "$template_dir/com.user.loop-scanner-watchdog.plist.template" > "$scanner_watchdog_plist"
+        if launchctl load "$scanner_watchdog_plist" 2>/dev/null; then
+            echo "[launchd] com.user.loop-scanner-watchdog loaded (every 5 min)"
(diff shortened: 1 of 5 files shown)


svv2014 (commented):
[loop:review] Clean, well-tested implementation of the heartbeat + watchdog path: `_write_heartbeat` integrates correctly at the top of each `run_once` tick, the watchdog handles the SIGTERM → SIGKILL escalation with a configurable grace period, and the launchd/cron registration is idempotent. The 7-test bats suite covers all the meaningful code paths (fresh/stale heartbeat, dry-run, DRY-RUN kill log). Minor nit: in `_kill_scanner`, the `|| return 0` inside the KILL_GRACE loop exits without executing `rm -f "$LOCK_FILE"` on the (normal) SIGTERM success path — the stale lock is benign because `lib/lock.sh` handles dead PIDs, but cleaning it up in the watchdog would be more explicit. Not a blocker given the existing stale-lock machinery. VERDICT: APPROVED — ready for QA.

svv2014:

PR #588, opened 2026-05-19
[LOOP-413] scanner: add liveness heartbeat and watchdog for silent-stop detection

svv2014:
Closes #413

## Changes

**`scanner/scanner.sh`** — writes a heartbeat file (`${LOOP_LOG_DIR}/scanner-heartbeat`) at the top of every `run_once()` tick. The file contains the current Unix epoch as a plain integer. The write is suppressed in `--dry-run` mode. This is the observable signal the watchdog consumes.

**`scanner/scanner-watchdog.sh`** (new) — a short-lived script that reads the heartbeat file's mtime, compares it to `LOOP_WATCHDOG_STALE_THRESHOLD` (default: `2 × LOOP_SCANNER_INTERVAL`, i.e. 10 min), and kills the scanner PID if stale. launchd's `KeepAlive=true` on the scanner plist then auto-restarts it within seconds. Supports `--dry-run` for safe testing. Designed to run every 5 min.

**`templates/launchd/com.user.loop-scanner-watchdog.plist.template`** (new) — launchd plist (`StartInterval 300`) for the watchdog. Installed by `--bootstrap` alongside scanner + reconciler.

**`install.sh`** — registers the watchdog in both `bootstrap_register_launchd()` (macOS) and `bootstrap_register_cron()` (Linux, `*/5` cron entry). Idempotent: skips if already registered.

**`tests/scanner-heartbeat.bats`** (new) — 8 bats tests covering: heartbeat file creation, epoch content, mtime progression, dry-run suppression, `run_once` integration, watchdog fresh/stale exit paths, and dry-run no-kill guard.

## Test Plan

- All 8 new bats tests pass locally (`bats tests/scanner-heartbeat.bats`)
- Manual verification: `kill -STOP $SCANNER_PID` → watchdog detects stale heartbeat within one 5-min interval → kills PID → launchd restarts

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -354,6 +354,25 @@ bootstrap_register_launchd() {
             echo "[launchd] WARNING: could not load com.user.loop-digest (check Console.app)" >&2
         fi
     fi
+
+    # Scanner liveness watchdog — checks heartbeat file every 5 min and kills
+    # a wedged scanner so launchd auto-restarts it via KeepAlive.
+    local scanner_watchdog_plist="$agents_dir/com.user.loop-scanner-watchdog.plist"
+    if [ -f "$scanner_watchdog_plist" ]; then
+        echo "[launchd] com.user.loop-scanner-watchdog already registered — skipping"
+    else
+        sed \
+            -e "s|__LOOP_ROOT__|$LOOP_ROOT|g" \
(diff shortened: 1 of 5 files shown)


svv2014:

